PR #1, opened 2021-10-25
Initial model setup

annehaley:
Basic creation of model files and a browseable API


marySalvi:
Looks good to me! Ran locally and confirmed browsable api is working properly

One thing is that master will need to be changed to main here:
https://github.com/atlascope/atlascope-api/blob/ea28236ae73a53a91a43febf471c199cfb25df58/.github/workflows/ci.yml#L4-L6

annehaley:
#2 was merged instead. This can be discarded.

PR #7, opened 2021-11-18
Slim models and robust API

annehaley:
Instead of merging #1, we should merge this instead. This branch fixes all that I learned from the first attempt.

This branch includes the following additions to `main` (which is currently just girder cookiecutter):
- Instead of 5 main models, we have three. All are initialized in the same migration file. We don't need `ContextMap` or `ConnectionsMap` to be explicit models, since they are one-to-one with Investigations, and we will never need to access them separately. Thus, this implementation has slimmer and simpler models.
- Instead of functional API views, this implementation used class-based API views with the DRF Viewset classes. This also has far fewer endpoints, just slimmed to basic getters of Users, Investigations, and Datasets. We will add more to these eight endpoints as the need arises, but the class-based views are more robust and flexible for additions.
- This implementation includes a beginner test suite, which comes with tests for each of the eight endpoints. We will add to the test suite as we add endpoints. Each endpoint is tested four times with different user permissions.

If this PR is approved and merged, we should discard #1.


marySalvi (approved):
Looks good to me!

PR #10, opened 2021-12-07
Importers

annehaley:
This PR introduces a structure for creating importer plugins for atlascope.  Available importer objects are determined by the plugins available to the current instance. Each of these plugins should define some function that accepts a source URI.

The dataset model now includes a CharField `importer` that refers to the unique name of an importer plugin (among the choices of what is installed). For example "vandy" refers to the installed plugin called `atlascope-vandy-importer`. Calling `some_dataset.perform_import()` will send the Dataset's `source_uri` to the import function defined in the plugin, which is referenced by the Dataset's `importer` value. There is now a REST endpoint to perform an import on a Dataset.

For development, we can put the source code for importer plugins in a subdirectory called `atlascope_plugins`, and the docker config will install them as packages in our dev environment.

For production, we should register these plugin repos with PyPI, and any production instances should `pip install` the plugin or include the PyPI dist name in `setup.py`.


annehaley:
^^ There's a lot of force-pushing above, mostly because I was working on fixing the CI tests (it was passing locally but not here). The first two commits contain the real changes.

atlascope/core/models/dataset.py
```diff
@@ -5,14 +5,18 @@
 from guardian.admin import GuardedModelAdmin
 from rest_framework import serializers
 
+from .importer import importers
+
+available_importer_choices = [(importer_name, importer_name) for importer_name in importers]
```
waxlamp: Why build a list of tuples like this? Is that a requirement for Django models to use a "choice of strings" kind of field (as in line 19 below)?

atlascope/core/rest/endpoints/dataset_endpoints.py
```diff
@@ -24,3 +28,18 @@ def get_queryset(self):
...
+    @object_permission_required(model=Dataset)
+    @action(
+        detail=True,
+        methods=['POST'],
+        url_path='import',
+    )
+    def perform_import(self, request, **kwargs):
+        dataset: Dataset = self.get_object()
+        dataset.perform_import()
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
waxlamp: Looks like we're not checking for the success of the import operation. I assume we'll want to add in some kind of error handling here. No need to add it now, but please file an issue to track that enhancement.

atlascope/core/rest/endpoints/dataset_endpoints.py
```diff
@@ -24,3 +28,18 @@ def get_queryset(self):
...
+
+    @swagger_auto_schema(
+        request_body=no_body,
+        responses={204: 'Import successful.'},
+    )
+    @object_permission_required(model=Dataset)
+    @action(
+        detail=True,
+        methods=['POST'],
+        url_path='import',
```
waxlamp, line 41: Can you tell me what `url_path` does? The docs on it are a bit sparse (keep in mind I'm not as good at Django as you all are).

atlascope_plugins/atlascope-vandy-importer/pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
```
waxlamp, line 3: Just curious: did this come out of a package generator type thing? If not, how did you know / decide to use `setuptools.build_meta` here?

atlascope_plugins/atlascope-vandy-importer/setup.py
```diff
@@ -0,0 +1,13 @@
+from setuptools import setup
+
+setup(
+    name="atlascope-vandy-importer",
+    version="0.0.1",
+    entry_points={
+        "atlascope.plugins": "vandy = import_function:run",
```
waxlamp, line 7: Ah, this is neat!

atlascope_plugins/atlascope-vandy-importer/setup.py
```diff
@@ -0,0 +1,13 @@
...
+    },
+    install_requires=[
+        "requests",
```
waxlamp, line 10: Why is `requests` listed as a requirement of this package?

atlascope_plugins/atlascope-vandy-importer/setup.py
```diff
@@ -0,0 +1,13 @@
...
+        'importlib; python_version == "2.6"',
```
waxlamp, line 11: Can you say what this line means?

dev/django.Dockerfile
```diff
@@ -18,3 +18,6 @@ RUN pip install --editable /opt/django-project[dev]
 
 # Use a directory name which will never be an import name, as isort considers this as first-party.
 WORKDIR /opt/django-project
+
+COPY atlascope_plugins /opt/atlascope-plugins/
+RUN pip install --editable /opt/atlascope-plugins/*
```
waxlamp, line 23: Does this mean that any change to an importer will require a Docker rebuild before the change can be seen in dev? If so, is there an easier way where we could specially treat the stuff in `atlascope_plugins` so that code edits could simply trigger a server restart, rather than requiring this rebuild? Would we lose the nice uniformity of having all such plugins go through entry points?

atlascope/core/migrations/0003_dataset_importers.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('core', '0002_base_models'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='dataset',
+            name='importer',
+            field=models.CharField(choices=[], max_length=100, null=True),
```
waxlamp: This migration says that the choice list is blank. Will that cause problems if we add or otherwise change the list of available importers at runtime? Or is that simply how choice-based Django model field work?

atlascope/tests/test_core_endpoints.py
```diff
@@ -140,6 +140,7 @@ def test_list_datasets(user_api_client, user, dataset_factory):
             'description': d.description,
             'source_uri': d.source_uri,
             'public': d.public,
+            'importer': d.importer,
```
waxlamp, line 143: Does [`dataset_factory`](https://github.com/atlascope/atlascope-api/blob/85a62c7174ca6d077d53a28bc8400ba6d19cf49a/atlascope/tests/factories.py#L17-L25) need to be updated to have a faked importer?

tox.ini
```diff
@@ -75,6 +78,8 @@ ignore =
     W503,
     # Missing docstring in *
     D10,
+    # Single vs double quotes conflicting with black
+    Q000,
```
waxlamp, line 82: Can you explain why this is needed? If we enable `flake8-black` or similar, would that help?

I'd prefer that our linting tools *not* conflict with Black, since that's our formatter of choice.

waxlamp (commented):
This looks great overall. Thanks for working on it. Just have a bunch of questions, and a couple of sticking points to resolve before I can approve this.

atlascope/core/models/dataset.py
annehaley: Yes, this is a Django thing: https://docs.djangoproject.com/en/4.0/ref/models/fields/#:~:text=A%20sequence%20consisting,name.%20For%20example%3A

atlascope/core/rest/endpoints/dataset_endpoints.py
annehaley: Adding a try-catch here is easy enough. I rolled this into 1ed50bbd1f9b953c6e3c6135c430de19fea2f6f4 before I realized it should be a separate commit :face_exhaling:

atlascope/core/rest/endpoints/dataset_endpoints.py
annehaley: Sure, it just changes the string of the rest endpoint URL. Without this argument, the `action` deocrator would automatically use the name of the function for the URL, and our endpoint would be `POST /api/vi/datasets/{id}/perform-import`. Since I can't make the function name itself `import` (as that is a reserved python word), I use this argument to make the endpoint simpler: `POST /api/vi/datasets/{id}/import`

atlascope_plugins/atlascope-vandy-importer/pyproject.toml
annehaley: I was using the `setuptools` quickstart guide for python packaging: https://setuptools.pypa.io/en/latest/userguide/quickstart.html

atlascope_plugins/atlascope-vandy-importer/setup.py
annehaley: Another copy-paste from the `setuptools` quickstart: https://setuptools.pypa.io/en/latest/userguide/quickstart.html#:~:text=setup.cfg-,setup.py,-from%20setuptools%20import

dev/django.Dockerfile
annehaley: Docker and pip actually work really well together on this. The `--editable` flag on a pip install command puts you in development mode on that package, and pip will watch for updates.

atlascope/core/migrations/0003_dataset_importers.py
annehaley: This part I am not so sure about. It worked fine for my setup, but I'm wondering if it will break on someone else's. If it does, we may have to find a complicated workaround, since each instance will have a different list of installed plugins. I think we will be safe, since most of the time, changes to the choices field don't warrant a full migration. Still, could you spin it up and give it a try?

atlascope/tests/test_core_endpoints.py
annehaley: yes, absolutely. Good catch.

tox.ini
annehaley: We already have `flake8-black` included in our `deps` for `testenv:lint`, and we can't invoke `flake8-black` directly, it works as a plugin to the `flake8` command. Strangely, even with `flake8-black` working, they seem to disagree on single and double quotes _only_ in a `setup.py` file. 

With this rule ignore, though, we are telling flake that we would like to trust black on this one.

atlascope/core/models/dataset.py
waxlamp: Ah I see. We will probably want to use the "human readable name" aspect of this in the future. Could you file an issue to divorce the importer name from a human-readable name?

dev/django.Dockerfile
waxlamp: How does pip watch for updates? And does that mean we *won't* need to do a Docker rebuild upon changing importer code?

annehaley: Yknow I'm glad you asked, because until now I was trusting something in the docs for `--editable` that I apparently misinterpreted. I _thought_ that this flag would take care of updates, but I suppose it doesn't work as well with docker as I thought.

It does take a build to get changes to the plugin. Perhaps we can figure out that special treatment in another PR? I filed #12 to track.

atlascope/core/models/dataset.py
annehaley: Not sure what you mean by divorce; any time we access the field, we will get the human-readable version by default. The "actual value" as Django calls it is simply what is saved in the db column. So If you have really long human-readable values that can be simplified to two-character codes, then there is no need to waste db space and you can set the max length of the field to 2 while the values you get back are much longer.

I'm not sure we can do much to avoid this structure for what `CharField` expects of the `choices` array...

atlascope_plugins/atlascope-vandy-importer/setup.py
annehaley: Another copy-paste from the setuptools quickstart: https://setuptools.pypa.io/en/latest/userguide/quickstart.html#:~:text=setup.cfg-,setup.py,-from%20setuptools%20import

When I pasted it in, I realized that I could take it out, but @marySalvi  most likely needs it for the functionality the `import_function` will perform.

atlascope/core/models/dataset.py
waxlamp: All I meant was, you're currently duplicating a single name into both values, and we might want to vary the names in the future. But TBH, this is not important enough to do anything further with at the moment, so I'll just resolve this thread. Thanks!

atlascope_plugins/atlascope-vandy-importer/setup.py
waxlamp: Seems odd since we don't even touch Python2 in this project. But it's it's not causing any harm, then we can move along. Might be good to figure out why they have you copypasta'ing this way (but not urgent at all).

dev/django.Dockerfile
waxlamp: I think `--editable` generally makes your installation immediately responsive to changes in the source files (since it is using symlinks or similar to "install" your package). But since we need to actually copy the files into the docker image before they can be pip-installed (or, something along these lines), we'll still need to rebuild the image. No worries though, just something we need to know about. Thanks for filing the issue.

atlascope/core/migrations/0003_dataset_importers.py
annehaley: Instead of dealing with possible issues on this choice field, @waxlamp pointed out that a validator function would be far easier, since its logic is done at runtime. This is a very elegant implementation that I had surprisingly never used! This change is made in ee3995f.

waxlamp: The change looks great. It's more explicit, and it avoids the semantic constraints inherent in the idea of the choice field.

waxlamp (approved):

PR #19, opened 2021-12-14
Bring frontend code into this repository

waxlamp:


web/tests/unit/example.spec.ts
```diff
@@ -0,0 +1,5 @@
+describe('Component under test', () => {
```
annehaley, line 1: Do we need this example test file at all? I can't tell what it does other than act as a template.

annehaley (changes requested):
This looks good, but I think we should merge #10 first and rebase this once `importers` is in `main`. It would solve the problem of the failing CI test here (because the `importers` branch specifies that we should stay on Django 3) and it would likely lead to less conflicts.

web/tests/unit/example.spec.ts
waxlamp: I think it's just there to fill out the testing infrastructure until actual tests are needed. It will go away when real tests arrive, and as such I think we do need it until then.

annehaley: valid enough reason to keep it :+1:

annehaley (approved):

PR #22, opened 2021-12-21
Create initial TS types for api calls and script for generating types from api schema

naglepuff:
This PR adds a new script that can be ran as a `yarn` command that generates typescript types corresponding to a swagger API. Information on how to run this script has been added to the front-end documentation. If running the `atlascope` back-end django project locally, the script can be run to generate types as follows:

`yarn run generateTypes http://localhost:8000`

This PR also includes an initial types file.

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -42,7 +42,7 @@ yarn run lint
 
 ### Generate Typescript Types
 
-If you make changes to the API for Atlascope, you can run the following command to regenerate types to be compatible with your changes.
+If you make changes to the API for Atlascope, make sure to run the following command to regenerate types to be compatible with your changes.
 ```shell
 yarn run generateTypes <Swagger schema URL>
 ```
--- a/web/README.md
+++ b/web/README.md
@@ -40,6 +40,14 @@ yarn run test:unit
 yarn run lint
 ```
 
+### Generate Typescript Types
+
+If you make changes to the API for Atlascope, you can run the following command to regenerate types to be compatible with your changes.
+```shell
+yarn run generateTypes <Swagger schema URL>
+```
+If you are running the API backend locally (see the `README.md` file at the root of this project for details), you can supply the URL `http://localhost:8000/api/docs/swagger`.
+
 ### Configuration
 
 This project is configured via environment variables. Reference the `.env` file for the available variables. The `.env.production` file contains additional variables relevant to production deployments.
--- a/web/package.json
+++ b/web/package.json
@@ -14,6 +14,7 @@
     "@vue/composition-api": "^1.0.0-rc.2",
     "axios": "^0.21.1",
     "geojs": "^1.4.3",
+    "swagger-typescript-api": "^9.3.1",
     "vue": "^2.6.11",
     "vue-router": "^3.2.0",
     "vuetify": "^2.4.0"
--- a/web/package.json
+++ b/web/package.json
@@ -6,7 +6,8 @@
     "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
     "test:unit": "vue-cli-service test:unit",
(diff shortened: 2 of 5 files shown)


web/README.md
```diff
@@ -40,6 +40,14 @@ yarn run test:unit
 yarn run lint
 ```
 
+### Generate Typescript Types
+
+If you make changes to the API for Atlascope, you can run the following command to regenerate types to be compatible with your changes.
```
waxlamp:
```suggestion
If you make changes to the API for Atlascope, make sure to run the following command to regenerate types to be compatible with your changes.
```

waxlamp (commented):
This is very cool!

mcovalt (approved):
This is great. I'm wondering if there is a test we can run on GitHub actions to make sure the types are up to date. I don't think that should block this from getting merged, but if you agree that would be helpful, I'll open up an issue.

waxlamp:
> This is great. I'm wondering if there is a test we can run on GitHub actions to make sure the types are up to date. I don't think that should block this from getting merged, but if you agree that would be helpful, I'll open up an issue.

Should just be able to build the project; any typechecking failures will cause the "test" itself to fail.

naglepuff:
> > This is great. I'm wondering if there is a test we can run on GitHub actions to make sure the types are up to date. I don't think that should block this from getting merged, but if you agree that would be helpful, I'll open up an issue.
> 
> Should just be able to build the project; any typechecking failures will cause the "test" itself to fail.

In the case that a PR makes certain changes to the back end only like adding a new data model, I imagine the build would still succeed. I think the developer of such a PR should also regenerate front-end types, to keep everything in sync. I might be worth looking into a github action to check for this, or maybe we could use a PR template (not sure if this exists or not) with a checklist as a friendly reminder.

mcovalt:
For instance, after https://github.com/atlascope/atlascope/pull/10, the generated types differed from what this branch initially provided. It would be nice for merging to be blocked unless https://github.com/atlascope/atlascope/pull/22/commits/607269d1781c25d8c01ff50dc9d73a8d006d4fec were applied.

waxlamp:
> > > This is great. I'm wondering if there is a test we can run on GitHub actions to make sure the types are up to date. I don't think that should block this from getting merged, but if you agree that would be helpful, I'll open up an issue.
> > 
> > 
> > Should just be able to build the project; any typechecking failures will cause the "test" itself to fail.
> 
> In the case that a PR makes certain changes to the back end only like adding a new data model, I imagine the build would still succeed.

You're right about this, good catch.

> I think the developer of such a PR should also regenerate front-end types, to keep everything in sync. I might be worth looking into a github action to check for this, or maybe we could use a PR template (not sure if this exists or not) with a checklist as a friendly reminder.

Let's not worry about this for right now. Having the types in the frontend is a big help for frontend development and that's reason enough to have them. As soon as someone encounters a problem at runtime where the types aren't matching up properly, they can regenerate the types (if the author of the breaking PR forgot to do so).

waxlamp:
> > > > This is great. I'm wondering if there is a test we can run on GitHub actions to make sure the types are up to date. I don't think that should block this from getting merged, but if you agree that would be helpful, I'll open up an issue.
> > > 
> > > 
> > > Should just be able to build the project; any typechecking failures will cause the "test" itself to fail.
> > 
> > 
> > In the case that a PR makes certain changes to the back end only like adding a new data model, I imagine the build would still succeed.
> 
> You're right about this, good catch.
> 

An easy way to write a test here would be to run the generator command and then `diff` the existing output and the new output. If there are any changes, `diff` returns nonzero, so you can use the return value to drive a test. We've used a similar technique on previous projects (Multinet maybe, but I can't quite remember).

PR #24, opened 2021-12-29
Add tiling endpoints

### LINKED ISSUE #5: set up a separate API in the backend for tiling endpoints
The titling API should be a sibling to the core API.

To quote @mcovalt's Slack message on this issue:
A "tile" is typically downloaded via a URL like this:
```
https://example.com/${z}/${x}/${y}.png
```
where

- x: an integer indicating the index of an image's tile where 0 is the left hand side of the image
- y: an integer indicating the index of an image's tile where 0 is the top of the image
- z: an integer indicating the "zoom level" where 0 is zoomed out all the way

This endpoint would typically return a 256px (tileWidth) by 256px (tileHeight) "tile".
The next zoom level will usually double the resolution of the previous. In other words, the tile returned by (x=0, y=0, z=0) will be composed of four tiles...

-  (x=0, y=0, z=1)
-  (x=1, y=0, z=1)
-  (x=0, y=1, z=1)
-  (x=1, y=1, z=1)

...like a region quadtree. Each of those are also 256px (tileWidth) by 256px (tileHeight) images. This forms the basis for being able to zoom very large images. Images are typically encoded with these tiles beforehand as scaling the image is too costly to do at each request.
At minimum, the frontend will need to know tileWidth and tileHeight, the range of z, and the ranges of x and y as a function of z.
The large_image package helps break apart a source image into tiles efficiently. Here's an example of a Django application using large image to serve tiles.

mcovalt:
Closes #5 
- [x] system environment parity: Ubuntu 20.04 is currently the Ubuntu LTS version used by Heroku and GitHub. While I imagine both of those services provide the ability to define these more specifically in a Docker container, Heroku's managed Postgres database uses Ubuntu 20.04's packaging for its PostGIS extension. I am anticipating we will use that PostGIS extension when we start using GeoDjango. Switching to Ubuntu LTS as the server host allows us to use the same libraries at runtime as these environments.
- [x] add `large_image[gdal]`: GDAL, GEOS, and PyPROJ are used with GeoDjango. The `large_image[gdal]` source works in tandem with these libraries and allows us to process a large number of image formats.
- [x] add tiling endpoints: we need to serve tiles for a [`Dataset`](https://github.com/atlascope/atlascope/blob/main/atlascope/core/models/dataset.py#L23-L41). GDAL can read remote images using its [Virtual File System interface](https://gdal.org/user/virtual_file_systems.html). This should be useful with `Dataset.source_uri`.


mcovalt:
The last few commits are my attempt to fix the GitHub Actions tests. Tox doesn't seem to be respecting the `--force-dep` option for GDAL. It is installing 3.4.0, not 3.0.4.

```bash
tox --force-dep gdal==3.0.4 --force-dep pyproj~=2.0
```

Mysteriously, this works for `pyproj`, though.

atlascope/core/rest/endpoints/tile_endpoints.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class TileMetadataView(GenericAPIView, mixins.RetrieveModelMixin):
+    queryset = Dataset.objects.all()
+    permission_classes = [IsAuthenticated]
+    serializer_class = TileMetadataSerializer
+
+    def get(self, *args, **kwargs):
+        dataset = self.get_object()
+        tile_source = GDALFileTileSource(dataset.source_uri)
```
mcovalt:
```suggestion
        tile_source = GDALFileTileSource(dataset.content)
```

When https://github.com/atlascope/atlascope/pull/25 lands.

atlascope/core/rest/endpoints/tile_endpoints.py
```diff
@@ -0,0 +1,96 @@
...
+                'y',
+                openapi.IN_PATH,
+                description='The 0-based y position of the tile on the specified z level.',
+                type=openapi.TYPE_INTEGER,
+            ),
+        ],
+    )
+    def get(self, *args, x=None, y=None, z=None, **kwargs):
+        dataset = self.get_object()
+        tile_source = GDALFileTileSource(dataset.source_uri, encoding='PNG')
```
mcovalt:
```suggestion
        tile_source = GDALFileTileSource(dataset.content, encoding='PNG')
```

When https://github.com/atlascope/atlascope/pull/25 lands.

atlascope/core/rest/endpoints/tile_endpoints.py
```diff
@@ -0,0 +1,96 @@
...
+        return Response(serializer.data)
+
+
+class TileSchemaGenerator(SwaggerAutoSchema):
+    def get_produces(self):
+        return ['image/png']
+
+
+class TileView(GenericAPIView, mixins.RetrieveModelMixin):
+    queryset = Dataset.objects.all()
```
mcovalt:
```suggestion
    queryset = Dataset.objects.filter(dataset_type='tile_source')
```

When https://github.com/atlascope/atlascope/pull/25 lands.

atlascope/core/rest/endpoints/tile_endpoints.py
```diff
@@ -0,0 +1,96 @@
...
+from rest_framework.generics import GenericAPIView
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+
+from atlascope.core.models import Dataset
+from atlascope.core.rest.additional_serializers import TileMetadataSerializer
+
+
+class TileMetadataView(GenericAPIView, mixins.RetrieveModelMixin):
+    queryset = Dataset.objects.all()
```
mcovalt:
```suggestion
    queryset = Dataset.objects.filter(dataset_type='tile_source')
```

When https://github.com/atlascope/atlascope/pull/25 lands.

mcovalt:
This PR is getting large in scope. I am going to split it up into separate PRs for the reviewer's sanity.

PR #25, opened 2022-01-04
Add to Dataset model

### LINKED ISSUE #13: Changes to Dataset model
The dataset model will need the following changes:

1. Addition of a JSONField to hold the metadata returned from the import function
2. Addition of an s3 File Field to hold the actual data file content
3. Addition of a `type` CharField to distinguish between tile source Datasets, tile overlay Datasets, analytics Datasets, etc.
4. Possible addition of a OneToManyField of `derived_datasets`?

annehaley:
Resolves #13.


marySalvi (approved):
Looks good to me. Definitely feels like a comprehensive dataset model

PR #26, opened 2022-01-05
Add Jobs

annehaley:
This PR makes an initial definition for the Job model to prove concept. This model may be revised later. This PR creates two example Jobs in the dataloader and spawns them.
Run `python3 manage.py populate` to test this. The outputs of the spawned jobs are printed to the console and saved to the Job object. 
The first job, Average Color, simply generates an image output, so its `outputs` array will look something like this: 
```
['/django-storage/e09ee116-d179-4066-b83d-ffad44a488f6_output_0.png']
```
To see this image, simply go to `localhost:9000/django-storage/e09ee116-d179-4066-b83d-ffad44a488f6_output_0.png`. We can work on rerouting minio access (port 9000) through Django (port 8000) later.

I would appreciate extra scrutiny on my implementation of manual communication with s3 storage in `tasks.py`


marySalvi (approved):
Very cool work. Looks good to me, was able to run locally.

PR #30, opened 2022-01-06
Fix dataset changes

### LINKED ISSUE #27: Cannot create a `Dataset` without specifying a `derived_dataset`
I'm not able to create a `Dataset` in the Django admin page without specifying a `derived_dataset`. Is `derived_dataset` intended to be a required field?

### LINKED ISSUE #28: `s3_file_field` related exception on Dataset admin page
The following exception is thrown when visiting the `Dataset` admin page:

```
's3_file_field' is not a registered namespace
```

I believe we need to [add the relevant URL](https://github.com/atlascope/atlascope/blob/01cee820f6be48b2484be18d6cfbdb91329d71bc/atlascope/urls.py#L36).

annehaley:
Resolves #27. Resolves #28.

### DIFF
--- a/atlascope/core/migrations/0004_new_dataset_fields.py
+++ b/atlascope/core/migrations/0004_new_dataset_fields.py
@@ -32,7 +32,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='dataset',
             name='derived_datasets',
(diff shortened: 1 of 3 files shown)


mcovalt (approved):
LGTM!

PR #31, opened 2022-01-07
Use Docker workflow with plugins

### LINKED ISSUE #12: See plugin changes without rebuild

mcovalt:
Closes #12

Changes to the plugins' source will be immediately reflected in the container running the debug server. This was accomplished by bind-mounting the plugin source directory into the Docker container via our Docker Compose configuration. An important caveat of this approach is that each plugin needs to be specifically installed in the Dockerfile.

### DIFF
--- a/dev/django.Dockerfile
+++ b/dev/django.Dockerfile
@@ -16,8 +16,8 @@ ENV PYTHONUNBUFFERED 1
 COPY ./setup.py /opt/django-project/setup.py
 RUN pip install --editable /opt/django-project[dev]
 
+COPY ./atlascope_plugins/atlascope-vandy-importer/setup.py /opt/atlascope-plugins/atlascope-vandy-importer/setup.py
+RUN pip install --editable /opt/atlascope-plugins/atlascope-vandy-importer
+
 # Use a directory name which will never be an import name, as isort considers this as first-party.
 WORKDIR /opt/django-project
-
-COPY atlascope_plugins /opt/atlascope-plugins/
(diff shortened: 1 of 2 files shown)


mcovalt:
This _should_ result in the debug server restarting, but I have not tested this; it depends on whether [Django's `runserver` recognizes that the plugins' source is supposed to be watched](https://github.com/django/django/blob/main/django/utils/autoreload.py#L290-L301).

annehaley:
So far, it appears that changes to the plugin files are still not recognized by Django without rebuild. Is there a way that we can add to the `watched_files` the linked file describes?

mcovalt:
Thanks for testing that. Does restarting the container show changes?

There are two things we are looking for here:
- Changes to the plugin's source files being seen in the Docker container
- Changes to the plugin's source files triggering a restart in Django's `runserver` process

I was experiencing the former not working until I did a `docker compose down`. Note that issuing that command will reset your database!

mcovalt:
I can confirm that:
- Changes to the plugin's source files ***are*** being seen in the Docker container
- Changes to the plugin's source files ***are not*** triggering a restart in Django's `runserver` process.

> Is there a way that we can add to the `watched_files` the linked file describes?

It doesn't seem we can easily do that by modifying any of the classes/methods in the linked module. I'm curious why it's not being seen. The server restarts even if we edit files in the system's `/usr/local/lib/python3.8/site-packages/...`.

annehaley (approved):

PR #32, opened 2022-01-07
Add dataset to dataloader

annehaley:
With recent testing of the `Dataset` model, it may be helpful if we are all using the same test objects. @marySalvi will be able to add to this object later.


mcovalt (changes requested):
The fixtures should be validated with our serializers during testing. Otherwise we risk having corrupt seed data when new business logic is expressed in the serializers.

atlascope/core/management/commands/populate.py
```diff
@@ -82,11 +93,15 @@ def command(password):
     for model, filename in MODEL_JSON_MAPPING:
         print('-----')
         objects = json.load(open(DATALOADER_DIR + filename))
-        for obj in objects:
-            obj, many_to_many_values, files_to_save, permissions = expand_references(obj, model)
-            db_obj = model(**obj)
+        for data_obj in objects:
+            # will raise Exception if invalid field or value in data
+            MODEL_SERIALIZERS[model](data=data_obj)
```
annehaley: Is this what you mean @mcovalt? This would prevent the creation of those objects if they cannot be instantiated as the proper model serializer.

mcovalt: Yes, didn't see this. Thanks for bringing it to my attention.

annehaley: Oh, you didn't miss it, don't worry! I just added it after your suggestion!

mcovalt: I see this was recently committed. Sorry for the confusion. Isn't the call to `.is_valid()` required? [See here](https://www.django-rest-framework.org/api-guide/serializers/#raising-an-exception-on-invalid-data).

annehaley: I attempted to put an incorrect field in one of the data objects and it correctly raised an exception preventing the save of that object and stopping the script. If we have `raise_exception=True` on the call to `is_valid`, wouldn't that give us the same behavior?

mcovalt: Not always. Validation in Django is spread across both database constraints and Django's application logic. [Not always in ways you'd think](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.ForeignKey.on_delete) [0]. [Django's default validation framework requires the explicit call to `.clean()`](https://docs.djangoproject.com/en/4.0/ref/models/instances/#validating-objects). This is used to handle custom validation outside the scope of what is possible with database constraints (and within the scope, too [1]).

The introduction of Django REST Framework further complicates this. [Django REST Framework never actually calls the `Model.clean()` method](https://www.django-rest-framework.org/community/3.0-announcement/#differences-between-modelserializer-validation-and-modelform). [Instead, it wants all validation to happen within its serializers](https://www.django-rest-framework.org/api-guide/validators/#validation-in-rest-framework)... kinda [2]...

Of course, there could be validation and business logic expressed in the view, too. Or middleware. But then we're really in trouble and have business logic all over the place. Instead, it should _all_ go in the serializer so we can more easily understand the constraints placed on our data.

**tl;dr**: we'll put more validation in the serializers later and a call to `.is_valid()` will run that validation.

[0] e.g. the `ON DELETE` constraint is never actually enforced in PostgreSQL. Instead, Django emulates this.
[1] e.g. unique constraints are tested by Django _and_ enforced by the database.
[2] Some constraints have to be put on the model for correct database representation (e.g. `unique=True` on a field). Django REST Framework's `ModelSerializer` inspects the model class on instantiation to pull in these constraints. See its [metaclass](https://github.com/encode/django-rest-framework/blob/c5be86a6dbf3d21b00a296af5994fa075826bf0b/rest_framework/serializers.py#L278-L340) and [its field building methods](https://github.com/encode/django-rest-framework/blob/c5be86a6dbf3d21b00a296af5994fa075826bf0b/rest_framework/serializers.py#L1199-L1630). It's terribly complex. The worst of it is that `ModelSerializer` is limited in capability and often you'll often need to convert a `ModelSerializer` class to a `Serializer` class. At that stage you need to explicitly add all this validation to each field. Getting that serializer's validation to match with Django's model validation is a complicated and error-prone process that is necessary due to these competing validation philosophies of Django, DRF, and Postgres. Another caveat, Django's Admin page relies on validation split between its forms and validation on the Django model and nothing to do with DRF.

mcovalt: It would be nice if this validation occurred during a test. That way, code that changes the constraints of our data will be blocked from merging if the fixtures are made invalid with those changes.

annehaley: That's a good point, thanks for explaining. I'll add `is_valid` and move the check to a tox test.

annehaley: Changed in 2e2738a

mcovalt: Great, thanks!

atlascope/tests/test_dataloader.py
```diff
@@ -0,0 +1,24 @@
...
+        clas.Meta.model: clas
+        for clas in atlascope_models.__all__
+        if issubclass(clas, serializers.ModelSerializer)
+    }
+    model_serializers[User] = UserSerializer
+
+    for model, filename in MODEL_JSON_MAPPING:
+        objects = json.load(open(DATALOADER_DIR + filename))
+        for data_obj in objects:
+            model_serializers[model](data=data_obj).is_valid()
```
mcovalt:
```suggestion
            assert model_serializers[model](data=data_obj).is_valid()
```

Otherwise, the test will still pass if `.is_valid() == False`.

mcovalt (changes requested):

atlascope/tests/test_dataloader.py
annehaley: of course! my bad, it was one of those instances where you quickly read a sentence and assume a word is there

mcovalt:
```suggestion
            model_serializers[model](data=data_obj).is_valid(raise_exception=True)
```

Alternatively, raise an exception. This might provide more meaningful test failure messages.

annehaley: Ah ok, I'll switch to that.

atlascope/tests/test_dataloader.py
```diff
@@ -0,0 +1,24 @@
...
+        clas.Meta.model: clas
+        for clas in atlascope_models.__all__
+        if issubclass(clas, serializers.ModelSerializer)
+    }
+    model_serializers[User] = UserSerializer
+
+    for model, filename in MODEL_JSON_MAPPING:
+        objects = json.load(open(DATALOADER_DIR + filename))
+        for data_obj in objects:
+            assert model_serializers[model](data=data_obj).is_valid()
```
mcovalt:
```suggestion
            model_serializers[model](data=data_obj).is_valid(raise_exception=True)
```

This might provide more meaningful test failures.

annehaley: Changed in 7186274. It appears it's already giving us a detailed error; it's catching on the dataset whose importer is "vandy" and the tox environment does not have our custom plugins installed. Do you know how to include this in our tox env?

mcovalt: > Do you know how to include this in our tox env?

I think so. I'll make a PR for that now.

mcovalt:
I've created a draft PR over at #44 with the testing related features I requested in this PR pulled out. Unfortunately, fixtures unrelated to the `Dataset` model are not passing validation. As a result, I see no reason to block this PR from getting merged.

@annehaley Do the commits I pulled out into #44 look like they cover the testing functionality you implemented in this branch? If so, can I reset this branch to set HEAD to the commit before adding serializer testing functionality (cd19b9b2b2dd717d8a9e4f4cc08dde2a5650f78d)?

annehaley:
Yes, #44 is a much better implementation. I'm glad you brought up this whole test idea, I didn't realize other fixtures wouldn't pass.

mcovalt (approved):

Issue #18, opened 2021-12-13
GeoDjango for storage of Pin locations


annehaley:

waxlamp:
Who is working on this one? Seems to be @mcovalt--could you give an update?

mcovalt:
GeoDjango is ready to be used. All that needs to be done is adding a [`PointField`](https://docs.djangoproject.com/en/3.2/ref/contrib/gis/model-api/#pointfield) to the [`Pin` model](https://github.com/atlascope/atlascope/blob/9f01330f770b53e2cbe1fe5fc644bdbe4d104d67/atlascope/core/models/pin.py#L23). I will make the PR.

waxlamp:
I'll close this because it's a design consideration within other features we're contemplating (so handled better as part of those issues).

PR #33, opened 2022-01-07
System environment parity

mcovalt:
In preparation for #5 and #18

Ubuntu 20.04 is currently the Ubuntu LTS version used by GitHub, Heroku, and Heroku's managed Postgres instance. [PostGIS](https://postgis.net) is an extension to PostgreSQL used by [GeoDjango](https://docs.djangoproject.com/en/4.0/ref/contrib/gis/). PostGIS and GeoDjango both depend on some scientific libraries:
- [GDAL](https://gdal.org)
- [PROJ](https://proj.org)
- [GEOS](https://libgeos.org)

The combination of these dependencies can have slightly different behaviors when different versions are used. We can reduce these inconsistencies by switching to Ubuntu 20.04 for our runtime environment.


mcovalt:
Ding ding ding!... https://github.com/atlascope/atlascope/pull/33/commits/86dea0be0aaefb4af4468bbddd952cb1eb1f5760

I'm going to clean up these commit messages.

annehaley (approved):

PR #34, opened 2022-01-07
Use Nginx to proxy requests to MinIO from within the Docker container

mcovalt:
Django knows that it should reach out to `minio:9000` to talk to MinIO when run in our Docker Compose configuration. We tell Django to link to `localhost:9000` for files hosted by MinIO since MinIO is attached to port 9000 of the host.

Django would not link to `localhost:9000` for these files in normal circumstances. Instead, it would link to a FQDN where the object store lives. That link is valid whether it is requested from within the Docker container or not.

However, in our development environment case, `localhost:9000` *is not* valid from within the container, but *is* valid from the host. This means that Django can't access the files that it itself is linking to.

This PR introduces a Nginx instance that is launched as part of the entrypoint to the Docker container. It listens locally on port 9000 and transparently proxies those requests to MinIO.


waxlamp:
How do other Girder 4 projects deal with this?

annehaley (approved):
I'll be interested to hear what you learn on Thursday about the best approach for this problem. For now, we'll put this in and hopefully gain more insight later.

PR #36, opened 2022-01-07
Install GeoDjango

mcovalt:
In preparation for kitware-resonant/cookiecutter-resonant#265 and kitware-resonant/django-composed-configuration#18

This PR installs [GeoDjango](https://docs.djangoproject.com/en/3.2/ref/contrib/gis/) along with the required system libraries. It updates the PostgreSQL service to include [PostGIS](https://postgis.net) and be a version similar to what Heroku provides.


atlascope/settings.py
```diff
@@ -25,6 +26,18 @@ class AtlascopeMixin(ConfigMixin):
...
+        environ_name='DATABASE_URL',
+        # django-configurations has environ_prefix=None by default here
+        environ_prefix='DJANGO',
+        environ_required=True,
+        # Additional kwargs to DatabaseURLValue are passed to dj-database-url,
+        # then passed through to the Django database options.
+        engine='django.contrib.gis.db.backends.postgis',
+        conn_max_age=600,
+    )
+
```
mcovalt, line 40: Can we make this change more precise with `mutate_configuration`?

tox.ini
```diff
@@ -62,7 +74,13 @@ passenv =
...
     dev
 deps =
-    importlib_metadata
+    # We explicitly reference `gdal` and `pyproj` here
+    # so that their versions can be constrained
+    # with tox's `--force-dep` option.
+    # This allows to us match the correct versions to what
+    # versions of library files are available.
+    gdal
+    pyproj
```
mcovalt, line 81: Can we install `numpy` before GDAL here? Does tox install deps in order?

atlascope/settings.py
```diff
@@ -59,4 +73,14 @@ class ProductionConfiguration(AtlascopeMixin, ProductionBaseConfiguration):
...
+    DATABASES = values.DatabaseURLValue(
+        environ_name='DATABASE_URL',
+        environ_prefix=None,
+        environ_required=True,
+        # Additional kwargs here.
+        engine='django.contrib.gis.db.backends.postgis',
+        conn_max_age=600,
+        # Heroku is expected to always provide SSL.
+        ssl_require=True,
+    )
```
annehaley, line 86: Same as above comment

setup.py
```diff
@@ -47,7 +47,7 @@
         'drf-yasg',
         'django-click',
         'django-guardian',
-        'importlib_metadata',
+        'importlib_metadata>=3.6',
```
annehaley, line 50: If you are able to figure out a better method for specifying the tox deps instead of using `sitepackages=True`, can this change be undone?

tox.ini
mcovalt: It seems all these dependencies are installed in one go, not one-by-one. Installing `numpy` before GDAL will not work. We still need `sitepackages = true`.

setup.py
mcovalt: No, this change should stay regardless. It was incorrectly specified before. The change to allow `systemsitepackages` uncovered this error. The system had an older version of `importlib_metadata` so the old dependency was satisfied.  However, we use functionality only available in version 3.6 and later. The server would not start with the older version.

atlascope/settings.py
mcovalt: I filed an issue upstream to see if this can be simplified: https://github.com/girder/django-composed-configuration/issues/161.

atlascope/settings.py
mcovalt: See https://github.com/atlascope/atlascope/pull/36#discussion_r783122449.

annehaley (approved):

PR #37, opened 2022-01-07
Add tiling endpoints

### LINKED ISSUE #5: set up a separate API in the backend for tiling endpoints

mcovalt:
Closes #5

This PR adds two new endpoints:
- `dataset/{pk}/tiles/metadata`: metadata related to the tiles of the `Dataset`
- `datasets/{pk}/tiles/{z}/{x}/{y}.png`: the specified tile of the `Dataset`

This uses just the system libraries necessary to build GeoDjango. Tiles are generated in a "streaming" fashion. The image is read remotely from the object store.

This was tested against the target Vandy image. I'm not sure how to include such a test or include seed data for such a large blob.


mcovalt:
Tests are failing due to tox configuration issues fixed in #35, namely https://github.com/atlascope/atlascope/pull/35/commits/e85a28b8ee97fac8edccb42d3d63e9c1f608fef9.

atlascope/core/rest/endpoints/tile_endpoints.py
```diff
@@ -0,0 +1,96 @@
...
+
+class TileView(GenericAPIView, mixins.RetrieveModelMixin):
+    queryset = Dataset.objects.filter(content__isnull=False, dataset_type='tile_source')
+    model = Dataset
+    permission_classes = [IsAuthenticated]
+
+    @swagger_auto_schema(
+        responses={200: 'Image file', 404: 'Image tile not found'},
+        auto_schema=TileSchemaGenerator,
+        manual_parameters=[
```
annehaley, line 53: The manual parameters look good and seem like sufficient documentation for this endpoint :+1:

atlascope/core/rest/endpoints/tile_endpoints.py
```diff
@@ -0,0 +1,96 @@
...
+    def get(self, *args, **kwargs):
+        dataset = self.get_object()
+        tile_source = GDALFileTileSource(f'/vsicurl/use_head=no&url={dataset.content.url}')
+        serializer = self.get_serializer(tile_source)
+        return Response(serializer.data)
+
+
+class TileSchemaGenerator(SwaggerAutoSchema):
+    def get_produces(self):
+        return ['image/png']
```
annehaley: You were right that the `@swagger_auto_schema` decorator provides no way to do this. Even with `extra_overrides`, the `SwaggerAutoSchema` class does not reference its own overrides in its definition of `get_produces`. You've done this the best way it can be done.

annehaley (approved):

atlascope/core/rest/endpoints/tile_endpoints.py
mcovalt: That's unfortunate. Thanks for double checking my work. I'll add a docstring to this class to explain it's reasoning.

atlascope/core/rest/endpoints/tile_endpoints.py
mcovalt: Great. Thanks!

atlascope/core/rest/endpoints/tile_endpoints.py
mcovalt: Docstring added in fd7b3f8f01845751c7d75a05f9cd68af70b14212.

PR #45, opened 2022-01-13
Tiling endpoint renderer

mcovalt:
This PR removes the need for the drf_yasg `TileSchemaGenerator`. It also correctly handles the `Accept: ` header for PNGs.


annehaley (approved):
I like this better :+1:

PR #50, opened 2022-01-17
Make Dataset source uri nullable

annehaley:


waxlamp:
@annehaley, if this closes an issue, could you reference it in the PR description?

mcovalt:
What does it mean if a `Dataset` has no `content` and no `source_uri`?

annehaley:
Hm, I think we want to prevent the creation of such Datasets. We can add a constraint that at least one of those fields is not null. I'll add it in this PR. [This](https://stackoverflow.com/questions/53085645/django-one-of-2-fields-must-not-be-null) suggests we do so in the `clean` method.

mcovalt:
We should not use the `clean` method for validation. It is only used for Django's Forms API which we will not use. Instead, the constraint should be expressed [as a database constraint](https://docs.djangoproject.com/en/4.0/ref/models/constraints/#checkconstraint). A [validation check](https://www.django-rest-framework.org/api-guide/validators/) may also be included in the serializer.

mcovalt (approved):
Looks good. Thanks for the changes! It really helps me understand the data model through these constraints.

PR #51, opened 2022-01-17
Create initial detail view for Investigations

### LINKED ISSUE #47: Implement basic InvestigationDetail view
This will have:
- navbar
- GeoJS-based map view
- sidebar
  - metadata about the investigation
  - list of pins associated therewith

naglepuff:
Closes #47

This PR adds the following changes:

- Update client-side typings from recent back end development
- Update store to include the axios instance from `main.ts`. Eventually this should likely be moved elsewhere
- Add `getters` and `getters.pins` to the store
- Add investigation title to toolbar when on the detail view
- Create initial `InvestigationDetail.vue`, including a map and sidebar
- Add some basic functionality to sidebar: display pins and metadata

Workflow walk through:

1. Follow the instructions in the various `README`s to set up an Atlascope front end and back end locally. Configfure OAuth so you can use the login functionality.
2. Run the `populate` command to get some data in to the system.
3. Navigate to `localhost:8080` (or wherever your front end is running) and login. Select any investigation from the initial list by clicking `Investigate`
4. You should be brought to the detail view for the investigation. The name of the investigation should now be displayed in the center of the App bar at the top of the page.
5. You should see a map view and a collapsed sidebar on the right hand side of the screen. Click the chevron to expand the sidebar.
6. The sidebar should contain two expansion panels: Pins and Metadata.
7. The pins panel shows two dummy pins (just their color and ID for now)
8. The metadata panel shows some metadata about the investigation.
9. Navigate back to the home page. The investigation name should be removed from the App bar.
10. Click a different investigation and repeat 4-8. The pins shown will be the same, but verify that the metadata is correct for the investigation you've chosen.

### DIFF
--- a/web/src/components/AtlascopeToolbar.vue
+++ b/web/src/components/AtlascopeToolbar.vue
@@ -14,6 +14,8 @@
         </router-link>
       </v-toolbar-title>
       <v-spacer />
+      <v-toolbar-title> {{ investigationName }}</v-toolbar-title>
+      <v-spacer />
       <v-btn
         v-if="userInfo"
         text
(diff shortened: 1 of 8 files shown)


mcovalt (approved):
This is great. I like moving the axios instance to the store for now.

PR #54, opened 2022-01-18
Rename all instances of "dataloader"

annehaley:
In the early stages of this project, I nicknamed the `populate` script as the dataloader, but this can be confusing. This PR renames all instances of "dataloader"


waxlamp (approved):
I was indeed confused by the terminology. Thanks for this quick fix!

PR #63, opened 2022-01-21
Vandy importer

### LINKED ISSUE #4: put example data from vandy into the `populate` script
The example data should be included as a Dataset object. This will help us determine what information we have and need to model, as well as information we lack and need to fill in.

### LINKED ISSUE #11: Write vandy importer code in atlascope-vandy-importer plugin
The code for importing a source uri for a dataset that uses the "vandy" importer should be in `atlascope_plugins/atlascope-vandy-importer/import_function.py`.

marySalvi:
Closes #4 and Closes #11

Adds Vandy Importer to Atlascope Plugin
Adds Vandy example data in Populate


README.md
```diff
@@ -15,6 +15,11 @@ This is the simplest configuration for developers to start with.
 3. Access the browseable API, starting at http://localhost:8000/api/v1/
 4. When finished, use `Ctrl+C`
 
+### Run Populate Script
+1. Get API Token from https://styx.neurology.emory.edu/girder/
+2. Run script:
+`docker-compose run -e DJANGO_API_TOKEN=[your token] --rm django ./manage.py populate`
+
```
annehaley, line 22: It's worth noting now that in the future we will likely need to change the way we pass in api keys to the import scripts. If we want to allow the user to create datasets down the road, it will need to be something we accept through the endpoint. This is fine for now, of course.

annehaley (approved):
These changes look great! Can't wait for it to hook into @naglepuff's work.

README.md
marySalvi: I made issue #64  for this so we don't lose track of this.

PR #65, opened 2022-01-24
Install importers in GitHub Actions

mcovalt:
It seems some versions of `setuptools` allow no package folder, but others don't? Moving `atlascope_vandy_importer` to its own package ensures it can be installed in GitHub Actions.


mcovalt:
@annehaley Feel free to close this if we are moving away from plugins for importers.

Issue #55, opened 2022-01-18
Add "create dataset" mechanism


waxlamp:
This would be something like a plus button in the frontend that would enable the user to get data into the system. The UI should support the importers present in the system. There should be one importer that allows for uploading a local file to S3.

annehaley:
It was also mentioned that it may be useful for the "create dataset" endpoint to automatically trigger the import function if the `importer` and `source_uri` fields are provided and the `content` field is null.

PR #69, opened 2022-01-27
Move importers

### LINKED ISSUE #67: Move importer functions to `core` app
We agreed that the simplest implementation for the importer functions would be regular pythonic functions within `core`, rather than a plugin architecture or a django apps architecture.

annehaley:
Resolves #67. Relates to #55


annehaley:
@marySalvi pointed out that the `request_body` serializer for the Create Dataset endpoint was confusing because some fields will always be non-applicable. Since we give the user the option to specify a dataset with either an upload or an importer, there are two separate potential `request_body` serializers we can allow (see 26ff616).

However, we can't offer two potential `request_body` serializers on the same endpoint. Should we recombine these to one (and allow the serializer to always have some non-applicable fields) or should we split the Create Dataset endpoint into two (one for import and one for upload)? Or is there some way to implement multiple `request_body` serializers that I haven't found?

annehaley:
Another idea: If we split the create endpoint in two, we would have an "upload dataset" endpoint and an "import dataset" endpoint. The "import dataset" endpoint would accept most of the normal fields plus some extra information that is not saved to the Dataset model, such as the importer to use and the arguments to pass it. Information relating to the import is only ever used in this endpoint, so why save it to the model once we've used it? Since we decided to remove the `source_uri` field from the Dataset, why not remove the `importer` field and `extension` field, too?

annehaley:
For both this PR and #70, The seven migration files we had have been squashed into one. To migrate your db, do the following in your django container:

```
python manage.py migrate core 0001
```

Then switch branches, then:

```
python manage.py migrate
```

This only needs to be done once.

waxlamp (approved):
This looks good, though I don't think this closes #55.

atlascope/core/models/dataset.py
```diff
@@ -56,10 +54,15 @@ def get_read_permission_groups():
...
-    def perform_import(self):
-        importer = importers[self.importer]
-        imported_content = importer(self.source_uri)
-        self.content.save(f'{self.name.replace(" ","_")}.{self.extension}', imported_content)
+    def perform_import(self, **kwargs):
+        importer = importers[self.importer]()
+        importer.run(**kwargs)
+
+        self.content.save(
+            f'{self.name.replace(" ","_")}.{self.extension}',
```
marySalvi: Is this working as expected? The remote appears to be using the name of the dataset and the local appears to be using the info provided by content.
Screenshot from 2022-02-03 12-25-30
Screenshot from 2022-02-03 12-25-10

In general I think it is best to use the value from content for 'local' and a name value supplied in meta for remote. I think it will help reduce duplicates

annehaley: I made `name` an optional argument in [acfca96](https://github.com/atlascope/atlascope/pull/69/commits/acfca9668306908d40d50b05cd2f46cfdf6b5a0c). 

This means when I provide a request body like the following:
```
{
  "importer": "VandyImporter",
  "import_arguments": {
    "file_id": "qwerty",
    "api_key": "123456789"
  }
}
```
The name of the resulting Dataset object is returned as "VandyImporter [id]" where [id] is the id of the Dataset object.

If "name" is passed into the request body, that is the name that will be used in the Dataset object.

We can also do something custom in the Importer's `perform_import` function itself by setting `self.dataset_name` to something we compute during the import.

marySalvi (approved):

PR #44, opened 2022-01-13
Test `populate` script with serializer

mcovalt:
Test related changes copied over from #32

The fixtures should be validated with our serializers during testing. Otherwise we risk having corrupt seed data when new business logic is expressed in the serializers.

PR #72, opened 2022-02-04
Add pytest for populate script

annehaley:
Replaces #44, now using the native importer definitions. Instead of testing each serializer on the objects written in the `json` files, we can just run the script. Using the serializers breaks on any referential fields.

Also, this PR makes setting DJANGO_API_TOKEN optional. If the env var is not set, then the imports that require it are skipped.

### DIFF
--- a/atlascope/tests/test_populate.py
+++ b/atlascope/tests/test_populate.py
@@ -1,5 +1,5 @@
-import os
 import pytest
+
 from atlascope.core.management.commands.populate import command as populate_command
 
 
--- a/atlascope/tests/test_populate.py
+++ b/atlascope/tests/test_populate.py
@@ -1,0 +1,8 @@
+import os
+import pytest
+from atlascope.core.management.commands.populate import command as populate_command
+
+
+@pytest.mark.django_db
+def test_populate_script():
+    populate_command()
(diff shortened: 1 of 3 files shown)


atlascope/tests/test_populate.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+from atlascope.core.management.commands.populate import command as populate_command
+
+
+@pytest.mark.django_db
+def test_populate_script():
+    populate_command()
```
waxlamp, line 8: As a general matter, I'm not sure how much this test adds. Running the populate script is a dev task being done all the time, so it's worth thinking about whether adding a test creates more burden on maintenance than just dealing with problems when they come up. In this case, the test is so short that it's probably a wash.

waxlamp (approved):
It looks like over and above adding the test, this PR actually fixes a bug, is that right?

atlascope/tests/test_populate.py
annehaley: True, but on @marySalvi's review of #69, she found several places where I had broken the populate script, and I hadn't realized because I don't run it at the end of developing a feature. Adding a test will ensure that we don't merge breaking changes.

annehaley:
> It looks like over and above adding the test, this PR actually fixes a bug, is that right?

No substantial bug fix, just accounting for edge cases of running the populate script with an invalid or missing api_key

mcovalt (approved):
Looks good. I'll close #44 when this gets merged.

PR #75, opened 2022-02-04
Move app-specific logic out of useGeoJS

### LINKED ISSUE #73: Remove application business logic from useGeoJS.ts
`useGeoJS.ts` should provide a convenient way to interact with the `geojs` library from our vue components. Initial development to show whole slide development put calls to the back-end API and Atlascope specific logic that should be handled by the vue components.

naglepuff:
Closes #73 

This PR separates API interaction out of `useGeoJS.ts` Components using `useGeoJS` should handle that kind of business logic. This is important to address now so the two don't get more intertwined down the road.

To test the changes, stand up Atlascope with the started data and verify that you can see the datasets in the `InvestigationDetail` view. Functionally nothing should have changed, this is a code style change.

### DIFF
--- a/web/src/utilities/useGeoJS.ts
+++ b/web/src/utilities/useGeoJS.ts
@@ -1,16 +1,14 @@
 import {
-  ref, onMounted, Ref, watch, computed,
+  ref, onMounted, Ref,
 } from '@vue/composition-api';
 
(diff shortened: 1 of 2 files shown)


mcovalt (approved):
Perfect. Thanks for making this change!

PR #79, opened 2022-02-12
Native job types

### LINKED ISSUE #68: Job Runs should create derived Dataset objects
When a Job Run is complete, a Dataset object of the type "derived" should also be created so that JobRun outputs can be used for further computation/visualization.

annehaley:
During this week's check-in, we decided we should define job types with native celery scripts instead of JobScript objects. However, there should still be an endpoint for listing available job types.

Additionally resolves #68 because these topics are interrelated


naglepuff:
Would there be a way to add the expected additional parameters to the response? Something like:

```
{
  "average_color": {
    "description": "Return the...",
    "additional_inputs": None
  },
  "brightest_n_pixels": {
    "description": "Return...",
    "additional_inputs": [{
      "name": "n",
      "type": "number",
    }]
  },
}
```

I don't think it should hold up this PR, since we're focusing on only two jobs for the moment, but in the future it would be nice for the client to have this information in some form or another.

annehaley:
> Would there be a way to add the expected additional parameters to the response? 

Do you mean the response for `GET /jobs/types`? If so, we can absolutely do that. It's actually a very small change and can totally be included on this PR

naglepuff:
> > Would there be a way to add the expected additional parameters to the response?

> Do you mean the response for GET /jobs/types? If so, we can absolutely do that. It's actually a very small change and can totally be included on this PR

Yes, that is what I meant. Awesome!

marySalvi (approved):
Looks great to me!

atlascope/core/job_types/__init__.py
```diff
@@ -0,0 +1,9 @@
+from . import average_color, brightest_n_pixels
+
+available_job_types = {
+    module.__name__.replace(module.__package__, '')[1:]: module.run
```
waxlamp: Is that `[1:]` meant to strip off a dot? If so, then I think this might be clearer:

```suggestion
    module.__name__.replace(f"{module.__package__}.", ''): module.run
```

but this feels pretty creaky too so let's just keep an eye on it.

A more solid but less automated way to handle this would be to manually list name/function pairs in this declaration. Perhaps fewer surprises to do it that way.

waxlamp (approved):
Looks great!

PR #82, opened 2022-02-17
Dataset embeddings

### LINKED ISSUE #81: `DatasetEmbedding` model
This Relationship model should allow us to specify how one Dataset is nested inside another within the context of an Investigation.

The following fields should be included: 

1. `context_investigation`
2. `parent_dataset`
3. `child_dataset`
4. `coordinates`

`coordinates` should be a JSON field which describes the coordinates of the child's upper left and lower right corners within the parent's coordinate space (in px). An example value could be as follows:
```
embedding_relationship.coordinates = [ [ 25, 25 ], [ 50, 50 ] ]
```

annehaley:
Resolves #81


atlascope/core/models/dataset_embedding.py
```diff
@@ -0,0 +1,42 @@
...
+        on_delete=models.CASCADE,
+        related_name='child_embeddings',
+    )
+    child = models.ForeignKey(
+        'Dataset',
+        on_delete=models.CASCADE,
+        related_name='parent_embeddings',
+    )
+    child_upper_left = PointField()
+    child_lower_right = PointField()
```
mcovalt: Can we combine these into a single bounding box? It will likely result in easier to write and more performant queries.

mcovalt: Perhaps we could enforce that the coordinates are indeed "`upper_left`" relative to "`lower_right`" in the meantime?

mcovalt: I think a change ought to be made before merging. Whichever is easier: the constraints or bounding box.

mcovalt (changes requested):
This is great! Glad to see GeoDjango in action. A small change request on the model constraints. See comment.

We'll likely need to approach the red/green/blue/yellow tiles differently in the future, but I see no reason to block the PR for that now. In my use of the red/green/blue/yellow tiles, I was emulating a very large image by simply returning one of those tiles regardless of the /z/x/y.png path. This gave the effect of uniform sharp noise at deep zooms up until the fake zoom limit, then blurry noise past that limit. I'm investigating another, more clear demo tile source in a future PR. I'll bring that up in our next meeting.

atlascope/core/models/dataset_embedding.py
annehaley: Switched to using a Polygon from bounding box in [8d2e8d5](https://github.com/atlascope/atlascope/pull/82/commits/8d2e8d590ed2a15e0274b50163ce62d9f1d91668). I'm thinking we can add a db constraint later when we're sure what constraints apply (e.g. should we allow shapes other than rectangles?)

mcovalt: > I'm thinking we can add a db constraint later when we're sure what constraints apply

Totally agree. This is what I need to move forward on the frontend.

> should we allow shapes other than rectangles?

Probably not, but I don't know yet. I will need to assume these polygons are rectilinear in order to prevent blowing up the complexity of the frontend. The frontend will break when this assumption breaks. We'll probably need to add that constraint in sooner-or-later. I'll open an issue after this gets merged.

mcovalt (approved):

atlascope/core/models/dataset_embedding.py
waxlamp: I know this is merged already, but there's no reason to worry about anything besides rectangles right now, especially if the rectangle assumption keeps things easy to work on for the rest of Phase 1.

PR #88, opened 2022-02-23
Pin locations

### LINKED ISSUE #87: Pins should give a location to a Dataset
The Pin model should have a foreign key to a parent dataset, foreign key to a child dataset, and point location of the child within the parent. Functionally, this will be similar to the Dataset Embedding model, except that the child dataset takes up no space within the parent and should be rendered as a dot rather than a tile image. Pins can also have preview images to show on hover and a color. Jobs that create datasets should also create Pins. 

@naglepuff and @mcovalt would this work?

annehaley:
The Datasets created by jobs have a pinned location inside the original dataset (upon which the job was run). Resolves #87.


waxlamp (commented):
Looks great to me. Deferring official approval to @naglepuff (he will be better able to comment on code quality, etc.).

atlascope/core/job_types/utils.py
```diff
@@ -10,3 +14,37 @@ def to_saveable_image(pil_image):
...
+        dataset_type='analytics',
+        source_dataset=original_dataset,
+    )
+    new_dataset.content.save(
+        f'{job_name.replace(" ", "_")}.png',
+        to_saveable_image(output_image),
+    )
+    new_dataset.save()
+
+    new_pin_location = [5, 5]
```
naglepuff, line 35: I'm assuming that this pin location is relative to its `parent_dataset`? I just wanted to clarify. 

Also, any reason to have it start at (5, 5) and not (0, 0)? 

Having multiple pins in the same location might not ultimately be an issue, but I'm fine with this approach for now.

naglepuff (approved):
Looks good to me. I had a couple of questions about pin location but nothing that I think should block this PR.

atlascope/core/job_types/utils.py
annehaley: Yes, these locations are relative to `parent_dataset`, which is the original dataset passed into the job. My thinking behind starting at [5, 5] was simply to avoid having pins on the edges of datasets. At [5, 5] it is clearly within the `parent_dataset`.

PR #89, opened 2022-02-23
Refactor models

### LINKED ISSUE #84: Provide way to link job results with Investigation
My understanding of performing analytics within Atlascope is that this is primarily done with the use of jobs, which will eventually need to be spawned by users from the front end.

This places these jobs within the context of an investigation, and therefore I believe there should be a way to obtain Jobs and job output datasets for a single investigation.

Right now we have `GET /jobs`. From this endpoint, I don't see a clear connection to job results or the investigation from which that job was launched.

We also have `GET /datasets`. Some datasets in the populate script are created by jobs, and this is reflected in the dataset's metadata. However it doesn't include the ID of the job, or the investigation from which this job was spawned.

I think these connections are all useful. Connecting a `Job` to its output (via new tables or more specific metadata) provides additional information about why those `Datasets` exist. Connecting the `Job` to the `Investigation` makes it easier to determine what the system should present to a user when looking at a specific `Investigation`.

@annehaley @waxlamp Interested in hearing your thoughts on this. Or if a solution already exists that I can't seem to find.

annehaley:
Resolves #84. This PR implements a model refactoring to match the following diagram and expose more relationships in the API.

Atlascope Class Diagram


atlascope/core/models/dataset.py
```diff
@@ -25,7 +25,10 @@ class Dataset(TimeStampedModel, models.Model):
         default=settings.DATASET_TYPES[0],
     )
     source_dataset = models.ForeignKey(
-        'Dataset', null=True, on_delete=models.PROTECT, related_name='derived_datasets'
+        'Dataset',
+        null=True,
+        on_delete=models.PROTECT,
```
waxlamp, line 30: Why `PROTECT`? If the source dataset is deleted, then it seems the `source_dataset` field of any derived datasets would become `null` (i.e., `SET_NULL` instead of `PROTECT`). What do you think?

atlascope/core/models/job.py
```diff
@@ -19,22 +19,49 @@ def validate_job_type(value):
...
+    )
     job_type = models.CharField(
-        default='average_color', max_length=100, validators=[validate_job_type]
+        default='average_color',
+        max_length=100,
+        validators=[validate_job_type],
+    )
+    original_dataset = models.ForeignKey(
+        'Dataset',
+        on_delete=models.CASCADE,
```
waxlamp, line 34: Feels like this should also be `SET_NULL`?

waxlamp (changes requested):
Looks good overall.

One general comment: I think I'd prefer if we just used the word `investigation` instead of `context`. That's clearer and more explicit about that particular relationship.

atlascope/core/models/dataset.py
naglepuff: `PROTECT` here prevents deleting datasets that are a `source_dataset` for some derived dataset, right?. I think this is the correct behavior, especially in the case where a dataset is used across multiple investigations.

atlascope/core/rest/endpoints/job_endpoints.py
```diff
@@ -20,9 +20,21 @@ class JobViewSet(
...
+            return JobSpawnSerializer
+        return JobDetailSerializer
+
+    def get_queryset(self):
+        if 'context' in self.request.GET:
+            investigation = Investigation.objects.get(
+                id=self.request.GET['context'],
+            )
+            return investigation.jobs.all()
+        return Job.objects.all()
```
naglepuff: I'm thinking that it might be more clear to have an endpoint `GET /investigations/{id}/jobs` instead of passing the investigation id as a query parameter.

naglepuff (changes requested):
Looks good. I had one suggestion about getting jobs for an investigation. Looks like this provides what is needed for working with jobs in the front end. FWIW I am in agreement with @waxlamp that I think `context` would be better named `investigation`.

atlascope/core/models/dataset.py
waxlamp: Yes, this is saying that if B is a dataset that was derived from A, then any attempts to delete B will fail because A holds a reference to it.

I don't think this is really correct behavior. I also am not sure it matters for us in Phase 1 (since we don't have a mechanism to delete datasets, nor do we really need one right now) but in general terms, it doesn't make sense to prevent deleting a dataset just because some other dataset wants to refer to it.

In developer terms, this restriction will make it harder for us to manually set things up via the admin console. I don't think we need to change it right now (and definitely not in this PR), but if it causes problems we should revisit.

atlascope/core/models/job.py
waxlamp: See thread above; we can leave this as is for now.

atlascope/core/rest/endpoints/job_endpoints.py
annehaley: Changed in [f476840](https://github.com/atlascope/atlascope/pull/89/commits/f47684059d947fe64e3cdccd0df1a23c015ca2bb)

naglepuff (approved):
Looks good to me! Thanks for making the changes.

waxlamp (approved):

PR #95, opened 2022-03-01
Jobs `complete` boolean field

### LINKED ISSUE #83: Job statuses
We need a way to fetch the status of a spawned job. 

Perhaps we could go as far as deleting the object once the run is complete? This would keep this table's storage limited to in-progress jobs, and the outputs and important information about a job are kept in the resulting Dataset that is created. Not sure how the frontend would reconcile this, though; this is just a thought.

### LINKED ISSUE #90: Show job status in admin console
I tried creating a job on the admin console, and when it shows up in the list there is no status listed. If the status were visible, we could meet the spirit of #58 for Phase I.

annehaley:
Rather than complicate the `status` field as a CharField with choices, it is easier just to use a BooleanField `complete`, since there are really only two distinguishable states for any Job.

Resolves #83. Resolves #90.


waxlamp (approved):
Looks good to me!

If during our Phase 1 work we have a situation where we need to know that a job failed (this is unlikely, but possible) then we'd want to bump this up to an enum field rather than a boolean.

PR #97, opened 2022-03-02
Remove auth

### LINKED ISSUE #94: Remove auth
Having users and auth in the Phase 1 prototype isn't needed, and it's been tripping us up on a few occasions, so let's just get rid of it.

waxlamp:
This PR removes auth stuff, including permissions checking, ownership of resources, and the concept of a user.

This simplifies testing and setup, and should allow for easier access to the codebase for smaller-time contributors like myself.

I need some good reviews of this to make sure I'm not breaking anything. In particular, I'm not sure if the populate script is working properly (I think it is, but I'm not sure), and I had to add an empty list `permission_classes` attribute to the Job model, even though it doesn't seem like I should need to do that.

Closes #94


atlascope/core/management/populate/investigations.json
```diff
@@ -1,45 +1,19 @@
 [
     {
         "name": "Whole Team Investigation",
```
annehaley, line 3: We should probably change the names of the datasets we create, and perhaps only create one.

atlascope/core/migrations/0009_auto_20220302_1642.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 3.2.12 on 2022-03-02 16:42
```
annehaley: Can you change the name of this migration file to be more specific? Like `remove_user_fields`

atlascope/urls.py
```diff
@@ -33,7 +30,6 @@
 
 urlpatterns = [
     path('accounts/', include('allauth.urls')),
-    path('oauth/', include('oauth2_provider.urls', namespace='oauth2_provider')),
```
annehaley, line 36: Perhaps we should rip out `oauth` entirely, since this is about simplification. At this point, we can remove `oauth2_provider` and `guardian` from `INSTALLED_APPS` in our settings so we don't end up migrating the db to support these things. Then we can also remove them from our dependencies.

atlascope/core/management/commands/populate.py
```diff
@@ -56,32 +46,15 @@ def expand_references(obj, model):
...
-        if field_name in obj:
-            permissions[permission] = [
-                User.objects.get(username=username) for username in obj[field_name]
-            ]
-            del obj[field_name]
-    return obj, many_to_many_values, files_to_save, permissions
+    return obj, many_to_many_values, files_to_save
 
 
-@click.option('--password', type=click.STRING, help='password to apply to all users')
```
annehaley, line 75: We need to update lines 8-9 in the README; the command invocation has changed and `createsuperuser` is now required if you want to use the admin console.

atlascope/core/management/commands/populate.py
```diff
@@ -2,44 +2,34 @@
...
-    (User, 'users.json'),
     (Dataset, 'datasets.json'),
     (Investigation, 'investigations.json'),
     (DatasetEmbedding, 'embeddings.json'),
     (Job, 'jobs.json'),
 ]
 
-DEFAULT_PASSWORD = 'letmein'
-
 DEFAULT_CLIENT_URI = 'http://localhost:8080/'
```
annehaley: we also don't need `DEFAULT_CLIENT_URI`

atlascope/core/management/commands/populate.py
```diff
@@ -56,32 +46,15 @@ def expand_references(obj, model):
             remote_model = found_many_to_many.remote_field.model
             if remote_model == Dataset:
                 many_to_many_values[field_name] = [remote_model.objects.get(name=x) for x in value]
-            elif remote_model == User:
-                many_to_many_values[field_name] = [remote_model.objects.get(email=x) for x in value]
+            # elif remote_model == User:
+                # many_to_many_values[field_name] = [remote_model.objects.get(email=x) for x in value]
```
annehaley: remove

atlascope/core/models/investigation.py
```diff
@@ -2,83 +2,32 @@
...
 class InvestigationSerializer(serializers.ModelSerializer):
-    owner = serializers.SerializerMethodField('get_owner')
-
-    def get_owner(self, obj):
-        return obj.owner.username or None
-
     class Meta:
         model = Investigation
-        fields = ('id', 'name', 'description', 'owner')
+        fields = ('id', 'name', 'description')
```
annehaley: we probably don't need the `InvestigationSerializer` vs `InvestigationDetailSerializer` distinction anymore, just use one serializer here

atlascope/core/rest/endpoints/job_endpoints.py
```diff
@@ -20,7 +19,7 @@ class JobViewSet(
 ):
     model = Job
     queryset = Job.objects.all().order_by('id')
-    permission_classes = [IsAuthenticated]
+    permission_classes = []
```
annehaley: You'll actually need this in the other viewsets as well. The reason that you were only getting 401s for job endpoints is because those tests are the only "write" endpoints we test. If we add a test for creating a dataset, the same thing would happen to that test without this line in `dataset_endpoints.py`

atlascope/core/rest/endpoints/pin_endpoints.py
```diff
@@ -8,14 +6,7 @@
 class PinViewSet(GenericViewSet):
     model = Pin
     serializer_class = PinSerializer
-    permission_classes = [IsAuthenticated]
 
     def get_queryset(self):
```
annehaley: You can probably do the same as you have done on other viewsets and simply specify `queryset = Pin.objects.all().order_by('id')` instead of `get_queryset`

annehaley (changes requested):
Sad to see so much code go, but it is looking a lot slimmer now. :smiling_face_with_tear: 
It appears there are a lot of unused imports now (see CI lint test). I didn't carefully read any of the changes in `web` (I defer to @naglepuff for those), but I did test using the frontend and it works fine without auth. The populate script works fine as well.

atlascope/core/migrations/0009_auto_20220302_1642.py
waxlamp: See 029dc08eb333596803b65493801f116dbe068b92.

atlascope/core/management/commands/populate.py
waxlamp: See 40f93d89f0845ea254775bca52d8d414f3bbe3a9.

atlascope/core/management/commands/populate.py
waxlamp: See b40c3f7ff058b2170bc7beb669320f731b72c3b8.

atlascope/core/management/commands/populate.py
waxlamp: I must have removed this between my PR going live and my seeing this comment; not sure which commit that was.

atlascope/core/models/investigation.py
waxlamp: See 74c829c5ae34fabd6b41d42a95591adb9ec0cd71.

atlascope/core/rest/endpoints/job_endpoints.py
waxlamp: Instead of doing that, I figured out the configuration option to disable permission management in all routes: fdedfe99cb8acab7b5e40a4e79364635b799d0a5.

atlascope/core/rest/endpoints/pin_endpoints.py
waxlamp: See 0458668a7bf17eac5a1eb723fcda1936121a3966.

atlascope/core/management/populate/investigations.json
waxlamp: Once @mcovalt's examples land in the codebase a bit more solidly, I think it would make sense to create a couple of investigations featuring the speckled test data and then a more realistic one, but I can take care of that in a future PR.

atlascope/urls.py
waxlamp: I agree in principle but also it won't do any harm to leave this stuff in place. The goal, more than simplification, is an easier or simplified developer experience. I looked into ripping these things out by the root and it seems like more trouble than it's worth (other than guardian, which I will do; stand by).

waxlamp: See 142f37f7e86129df41dedef41bda1f42b6727b92.

waxlamp:
> Sad to see so much code go, but it is looking a lot slimmer now. smiling_face_with_tear

Like the Klingons, I do not consider the occasion of code deletion to be one of mourning, but rather celebration, as it rises to the ranks of the honored deleted in [Suto'vo'qor](https://memory-alpha.fandom.com/wiki/Sto-vo-kor). If you prefer human reflections, [Shakespeare](https://www.poetryfoundation.org/poems/45098/sonnet-71-no-longer-mourn-for-me-when-i-am-dead) is surprisingly relevant 😮 

But either way, this is ready for another look 😃 

@naglepuff, still need your eyes on the frontend changes.

annehaley (approved):

naglepuff (approved):
Looks good to me! Client side code changes were straightforward and I was able to do some testing with the front end app and nothing appears to be broken. Thanks for doing this. I guess we don't technically need the `girder-oauth-client` library anymore but I think you mentioned in another comment that there's not really any harm in leaving those kinds of things in for now.

waxlamp:
> Looks good to me! Client side code changes were straightforward and I was able to do some testing with the front end app and nothing appears to be broken. Thanks for doing this. I guess we don't technically need the `girder-oauth-client` library anymore but I think you mentioned in another comment that there's not really any harm in leaving those kinds of things in for now.

Frontend libraries are easier to remove than the backend ones, just because a lot of the default Django settings refer to those. I'll rip out girder-oauth-client.

PR #98, opened 2022-03-02
Fix migration 7 to work with non-empty database

annehaley:


waxlamp (commented):
Cool! How do I apply this if I've already "gone past" this migration in my own dev environment?

annehaley:
> Cool! How do I apply this if I've already "gone past" this migration in my own dev environment?

If you have some migration file on another branch that you've applied to your db, you'll need to do the following:
1. checkout the ahead branch
2. unapply the ahead migrations while those migration files are present (with `manage.py migrate core 000x` where 000x is the last shared migration between the branches)
3. checkout the behind branch

But that's only if you really want to try it out. If you're already ahead, you don't need to go back and apply this migration with these changes. The only difference is the value `migrate` will use to populate rows when it is adding the `investigation` columns to those three tables

waxlamp (approved):

Issue #74, opened 2022-02-04
Enable creation of subimage datasets from the InvestigationDetail view


naglepuff:
Since one recurring idea for Phase 1 has been selecting regions of an image and running computations on them, users should be able to somehow select these regions of an image. 

This could look like:
1. User draws a rectangle feature on the image. The user could be prompted to run a job with the rectangle coordinates as inputs
2. User chooses a job that runs on a subimage. Whatever portion of the whole slide image is visible in the viewer is used as the subimage

waxlamp:
We're going more with (1), but deferring the mechanism for running a job. Perhaps we can put together a modal dialog for launching a job on an image or subimage, but let's wait until some of this issue is settled first.

annehaley:
Resolved by #127

PR #99, opened 2022-03-02
add endpoint for creating subimage

marySalvi:
This creates the endpoint needed for the first steps for #74 

**Please note this PR also updates tox configuration for linting and black**


atlascope/core/rest/endpoints/dataset_endpoints.py
```diff
@@ -44,3 +50,21 @@ def create(self, request):
...
         return Response(DatasetSerializer(new_dataset_obj).data, status=status.HTTP_201_CREATED)
+
+    @swagger_auto_schema(request_body=DatasetSubImageSerializer())
+    @action(detail=False, methods=['POST'])
+    def createSubImage(self, request):
+        serializer = DatasetSubImageSerializer(data=self.request.data)
+        serializer.is_valid(raise_exception=True)
+        original_dataset = Dataset.objects.get(id=serializer.validated_data["original_dataset_id"])
+        meta = serializer.validated_data
+        meta.pop("original_dataset_id")
```
marySalvi, line 48: This part confuses the heck out of me. It appears here that removing the `original_dataset_id` from meta also affects the `serializer.validated_data`. Can someone help me understand what is happening here? @waxlamp @annehaley

waxlamp: My first impression is that `serializer.validated_data` is a mutable type, so `meta = serializer.validated_data` simply creates an alias. When you `meta.pop()`, you're mutating `meta` which also mutates `serializer.validated_data` (since those are just two names for the same underlying object).

Depending on what you need to do here, you could figure out how to make a copy of the data represented by `validated_data`, and then modify that.

marySalvi: Thanks! You were right. I think I have it correct now.

atlascope/core/migrations/0009_alter_dataset_dataset_type.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 3.2.12 on 2022-03-03 14:46
```
waxlamp: I think this will end up conflicting with my remove-auth branch (that one also has a 0009 migration). A rebase and cleanup should help to fix the problem.

atlascope/core/models/dataset.py
```diff
@@ -51,6 +51,9 @@ def perform_import(self, importer="UploadImporter", **kwargs):
         if not self.name:
             self.name = importer_obj.dataset_name or f'{importer} {self.id}'
 
+    # def create_sub_image_DS(self):
+    #     self.content.save(self.)
+
```
waxlamp:
```suggestion
```

Remove this?

atlascope/core/models/dataset.py
```diff
@@ -115,6 +118,18 @@ def create(self, validated_data):
...
+class DatasetSubImageSerializer(serializers.Serializer):
+
+    original_dataset_id = serializers.UUIDField(required=True)
+    x0 = serializers.IntegerField(required=True)
+    y0 = serializers.IntegerField(required=True)
+    x1 = serializers.IntegerField(required=True)
+    y1 = serializers.IntegerField(required=True)
+
+    def validate_original_dataset_id(self, value):
+        return str(value)
```
waxlamp: I think this should validate that `original_dataset_id` actually refers to a dataset.

atlascope/core/rest/endpoints/dataset_endpoints.py
```diff
@@ -44,3 +50,21 @@ def create(self, request):
         )
 
         return Response(DatasetSerializer(new_dataset_obj).data, status=status.HTTP_201_CREATED)
+
+    @swagger_auto_schema(request_body=DatasetSubImageSerializer())
+    @action(detail=False, methods=['POST'])
+    def create_sub_image(self, request):
```
waxlamp: Following RESTful style this should just be called `subimage`. Then the REST route will be `POST /datasets/{id}/subimage` which we could take to mean "create a subimage dataset from dataset {id}".

atlascope/core/rest/endpoints/dataset_endpoints.py
waxlamp: A warning: `copy()` does a shallow copy, and that may be sufficient for this case (if `validated_data` only has POD ("plain old datatype") values in it), but in general you would want `deepcopy()`. (See the docs on the [`copy`](https://docs.python.org/3/library/copy.html) module.)

atlascope/core/rest/endpoints/dataset_endpoints.py
```diff
@@ -44,3 +50,21 @@ def create(self, request):
...
+        serializer = DatasetSubImageSerializer(data=self.request.data)
+        serializer.is_valid(raise_exception=True)
+        original_dataset = Dataset.objects.get(id=serializer.validated_data["original_dataset_id"])
+        meta = serializer.validated_data.copy()
+        meta.pop("original_dataset_id")
+        new_dataset_obj = Dataset(
+            name=f'{original_dataset.name} Sub Image',
```
waxlamp:
```suggestion
            name=f'{original_dataset.name} Subimage',
```

atlascope/core/rest/endpoints/dataset_endpoints.py
```diff
@@ -44,3 +50,21 @@ def create(self, request):
...
+            metadata=meta,
+            source_dataset=original_dataset,
+            dataset_type="sub_image",
```
waxlamp:
```suggestion
            dataset_type="subimage",
```

atlascope/core/rest/endpoints/dataset_endpoints.py
waxlamp: In the future, we may find it useful to be able to override this name. No need to put in that feature now.

atlascope/settings.py
```diff
@@ -24,6 +24,7 @@ class AtlascopeMixin(ConfigMixin):
         'tile_source',
         'tile_overlay',
         'analytics',
+        'sub_image',
```
waxlamp:
```suggestion
        'subimage',
```

pyproject.toml
```diff
@@ -2,7 +2,7 @@
 line-length = 100
 skip-string-normalization = true
 target-version = ["py38"]
-exclude='\.eggs|\.git|\.mypy_cache|\.tox|\.venv|_build|buck-out|build|dist'
+exclude='\.eggs|\.git|\.mypy_cache|\.tox|\.venv|_build|buck-out|build|dist|web'
```
waxlamp, line 5: Nice. This will conflict with a different fix I have in the remove-auth branch, but yours is better. I'll remove the fix from my branch and rely on this one instead.

(A note: we both did a bad thing by including this fix in our feature branches; better would be to cherry pick this out to a PR against main so that everyone could have benefited from our genius independently of these big feature branches we're needing to review before merge.)

tox.ini
```diff
@@ -97,7 +97,16 @@ ignore =
...
+black-config = pyproject.toml
+exclude =
+    .git,
+    __pycache__,
+    .tox,
+    .env,
+    .venv,
+    env,
+    venv,
+    */node_modules
```
waxlamp: Why add this here if pyproject.toml is configuring black already?

waxlamp (changes requested):

atlascope/core/migrations/0009_alter_dataset_dataset_type.py
marySalvi: should be fixed

waxlamp (approved):

tox.ini
annehaley: I think undoing the changes to this file should fix the CI tests, since they are throwing the error "duplicate key exclude"

annehaley (approved):

tox.ini
marySalvi: the issue with the 'duplicate key exclude' was an uncaught duplicate from Roni's change. It has been fixed. I did remove the reference to pyproject.toml because it worked either way.

annehaley (changes requested):
Can you also add a pytest for your new endpoint?

marySalvi:
Pytest added. @annehaley

annehaley (approved):

tox.ini
marySalvi: @waxlamp  18e1f9fca0e18439fe96718fa219d7fa8c5b9093

PR #100, opened 2022-03-03
Add a management command to dump the full django settings

waxlamp:
I jumped ahead and put this small management script together without looking too hard at whether there's a baked-in way to do it. Take a look and let me know what you think.

### DIFF
--- a/atlascope/core/management/commands/showconfig.py
+++ b/atlascope/core/management/commands/showconfig.py
@@ -1,0 +1,38 @@
+from pprint import pprint
+
+import django.conf
+import djclick as click
+
+
+def remove_hidden_keys(d: dict) -> dict:
+    """
+    Recursively remove hidden keys from a dict.
+
+    For django settings, "hidden" keys would be ones that either (a) begin with
+    an underscore or (b) are not all-uppercase.
+    """
+    # You can't remove keys from a dict during iteration, so instead we collect
+    # the keys to be removed in a list and do them in a batch after the
+    # iteration pass.
+    remove = []
+    for k in d:
+        if k.startswith('_') or k.upper() != k:
+            remove.append(k)
+        elif type(d[k]) is dict:
+            remove_hidden_keys(d[k])
+
+    for k in remove:
+        del d[k]
+
+    return d
+
+
+@click.command()
+def command():
+    settings = django.conf.settings
+    config_dict = {}
+
+    for key in dir(settings):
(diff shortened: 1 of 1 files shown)


annehaley (approved):
I tried it and it works pretty well. :+1: However, if you're considering making this an upstream change for girder 4, perhaps a verbose vs quiet option for this command would be useful? This gives a lot of printout, and in many cases, I won't care about django default settings that haven't been overridden.

waxlamp:
> I tried it and it works pretty well. +1 However, if you're considering making this an upstream change for girder 4, perhaps a verbose vs quiet option for this command would be useful? This gives a lot of printout, and in many cases, I won't care about django default settings that haven't been overridden.

I'm not sure how I could arrange to only show things that have been changed, since the way django-configurations works is to instantiate a class, run some hooks, and then slam all the uppercase attributes into the live config. Do you have any ideas?

annehaley:
The [docs](https://docs.djangoproject.com/en/4.0/topics/settings/#:~:text=A%20Django%20settings%20file%20doesn%E2%80%99t%20have%20to%20define%20any%20settings%20if%20it%20doesn%E2%80%99t%20need%20to.%20Each%20setting%20has%20a%20sensible%20default%20value.%20These%20defaults%20live%20in%20the%20module%20django/conf/global_settings.py) say that the default settings are defined in `django.conf.global_settings.py`, so you could compare the output dict to the following:

```
from django.conf import global_settings

defaults = global_settings.__dict__
```

waxlamp:
> The [docs](https://docs.djangoproject.com/en/4.0/topics/settings/#:~:text=A%20Django%20settings%20file%20doesn%E2%80%99t%20have%20to%20define%20any%20settings%20if%20it%20doesn%E2%80%99t%20need%20to.%20Each%20setting%20has%20a%20sensible%20default%20value.%20These%20defaults%20live%20in%20the%20module%20django/conf/global_settings.py) say that the default settings are defined in `django.conf.global_settings.py`, so you could compare the output dict to the following:
> 
> ```
> from django.conf import global_settings
> 
> defaults = global_settings.__dict__
> ```

Cool, thank you! I will try something out with this.

waxlamp:
Of course this already exists. This command:

```
docker-compose run --rm django python ./manage.py diffsettings --output unified --all
```

does a very nice job of showing all settings together with the diff against the global defaults. Very handy, wish I knew about it before :)

PR #101, opened 2022-03-08
GeoJS projections

### LINKED ISSUE #93: Make GeoJS map from list of `DatasetEmbeddings`
A fixed representation of dataset relationships are demonstrated in the branch [`geojs-projections`](https://github.com/atlascope/atlascope/tree/geojs-projections). This issue tracks the progress for making this GeoJS map dynamically as a function of an `Investigation`'s list of `DatasetEmbedding`'s.

mcovalt:
Closes #93 

- [x] Deprecate `datasets.ts` @naglepuff
  - [x] Generate types from #106
  - [x] `datasets.getEmbeddings` should be replaced with a call to the API. It should be added to the list of promises fetched on investigation load.
  - [x] `datasets.getTileMetadata` should be replaced with a call to the API. It should be added to the list of promises fetched on investigation load.
- [x] Deprecate `MapProjections.vue` @mcovalt 
  - [x] Remove `/demo` route
  - [x] Isolate GeoJS commands to `useGeoJS.ts`
- [x] Change populate script to only include real datasets.
- [x] Revert ce81498cb0ad0c769c0387e02428710d4dc87d38 when #106 is merged
- [x] Revert ec6cb2601e188294c7e7fff2afcae6456d0dae18 when #106 is merged
- [x] Revert 56e019dcd362c9430aced18f9c27fc19ab9fc3f1 when #107 is merged

To test:
- go to the "Whole Team Investigations" embeddings link
- Zoom in on the top left of the image
- An investigation without embeddings will not display a map

https://user-images.githubusercontent.com/13244041/159101724-e027d953-eacd-4e5d-baec-ea27b0c7f8ad.mp4


web/.env
```diff
@@ -1,3 +1,3 @@
 VUE_APP_API_ROOT=http://localhost:8000/api/v1
 VUE_APP_OAUTH_API_ROOT=http://localhost:8000/oauth/
-VUE_APP_OAUTH_CLIENT_ID=CHANGEME
+VUE_APP_OAUTH_CLIENT_ID=cBmD6D6F2YAmMWHNQZFPUr4OpaXVpW5w4Thod6Kj
```
mcovalt: This should be reverted.

web/src/views/MapProjection.vue
```diff
@@ -0,0 +1,206 @@
...
+  setup() {
+    const mapElement = ref(null);
+
+    onMounted(async () => {
+
+      const embeddings: Array<DatasetEmbedding> = getEmbeddings(2);
+      const rootDatasetID: DatasetID = embeddings.find(e => embeddings.every(x => x.child !== e.parent)).parent;
+      const rootTileMetadata = getTileMetadata(rootDatasetID);
+      const rootMinLevel = Math.min(0, Math.floor(Math.log2(Math.min(500 / rootTileMetadata.tile_size, 500 / rootTileMetadata.tile_size))));
+      const rootMaxLevel = Math.ceil(Math.log2(Math.max(rootTileMetadata.size_x / rootTileMetadata.tile_size, rootTileMetadata.size_y / rootTileMetadata.tile_size)));
```
mcovalt: The `500`'s here are based on the size of `#map` DOM node. They are hard-coded, but ought to be found from the `mapElement` value.

web/src/views/MapProjection.vue
```diff
@@ -0,0 +1,206 @@
...
+          right: rootTileMetadata.size_x,
+          bottom: rootTileMetadata.size_y,
+        },
+        unitsPerPixel: Math.pow(2, rootMaxLevel),
+        center: {
+          x: rootTileMetadata.size_x / 2,
+          y: rootTileMetadata.size_y / 2,
+        },
+        min: rootMinLevel,
+        max: rootMaxLevel + 10,
```
mcovalt: The `rootMaxLevel + 10` is arbitrary. It's saying "get me 10 zoom levels past the bare minimum needed to zoom into the root fully". We're hoping that any child dataset doesn't need more that 10 extra zoom levels. We should rather just determine what the maximum zoom level is and change these parameters in the map.

web/src/views/MapProjection.vue
```diff
@@ -0,0 +1,206 @@
...
+export default defineComponent({
+  name: 'MapProjection',
+
+  setup() {
+    const mapElement = ref(null);
+
+    onMounted(async () => {
+
+      const embeddings: Array<DatasetEmbedding> = getEmbeddings(2);
+      const rootDatasetID: DatasetID = embeddings.find(e => embeddings.every(x => x.child !== e.parent)).parent;
```
mcovalt: This could be `undefined` meaning: no root node. It could also pick the first root node of many. We ought to implement backend constraints to prevent either case.

web/src/views/MapProjection.vue
```diff
@@ -0,0 +1,206 @@
...
+            tilesMaxBounds(level: number) {
+              return {
+                x: Math.floor( tileMetadata.size_x / Math.pow(2, rootMaxLevel - level) ),
+                y: Math.floor( tileMetadata.size_y / Math.pow(2, rootMaxLevel - level) ),
+              };
+            },
+          }
+        );
+        layer.gcs(`+proj=longlat +axis=enu +xoff=-${offset.x} +yoff=${offset.y} +s11=${1 / scale} +s22=${1 / scale}`);
+        const frontier = embeddings.filter(e => e.parent === datasetID);
```
mcovalt: This should check for cycles.

mcovalt:
@naglepuff We'll also need the tile metadata for each parent and child of each embedding, too.

Edit:

I think I've got it. Could you review these commits?
- a094fd430cbce54bf6efb6d5d9f4a8e5f5fb231d
- 571fcb78e745c819ab542f6fd7c85d785df41767

web/src/store/index.ts
```diff
@@ -94,6 +106,45 @@ const {
...
+          });
+          embeddings.forEach((embedding) => {
+            const promise = store.state.axiosInstance?.get(`/datasets/${embedding.parent}/tiles/metadata`).then((result) => ({
+              datasetId: embedding.parent,
+              result,
+            }));
+            if (promise) {
+              metadataPromises.push(promise);
+            }
+          });
```
naglepuff, line 142: Any reason not to create both promises in the same iteration over `embeddings`?

naglepuff: Just kidding.

mcovalt: Just bad code. One does `embedding.parent` and the other does `embedding.child`.

web/src/store/index.ts
```diff
@@ -94,6 +106,45 @@ const {
...
+          tileSourceDatasets.forEach((dataset) => {
+            const promise = store.state.axiosInstance?.get(`/datasets/${dataset.id}/tiles/metadata`).then((result) => ({
+              datasetId: dataset.id,
+              result,
+            }));
+            if (promise) {
+              metadataPromises.push(promise);
+            }
+          });
+          embeddings.forEach((embedding) => {
```
naglepuff, line 125: So with your recent changes, we're retrieving and storing the metadata for all `Datasets` that are:

1. Related to the investigation through the `Investigation.datasets` relation an of type `tile_source`
2. Related to the investigation by being either the `parent` or `child` property of an `Embedding`

Since a `Dataset` can be included in both (1) and (2), I think there's some redundancy with the current approach. I'm fine to leave it as is for now, since it works, but making a set of `Dataset` ids and then building the promise array from that might be a cleaner option.

I imagine that all Datasets in (2) should also be in (1), but I don't think our DB enforces that, and its possible that my interpretation of our data model is slightly off. But that's a separate discussion.

mcovalt: > Since a `Dataset` can be included in both (1) and (2), I think there's some redundancy with the current approach. I'm fine to leave it as is for now, since it works, but making a set of `Dataset` ids and then building the promise array from that might be a cleaner option.

Totally agree, but, I think that array of all datasets is used for the dropdown selector in the original view. I was afraid adding the `Dataset` IDs related to the embeddings to that array would have unintended consequences.

naglepuff: I've opened #110 to capture my perspective on investigation/dataset relationships. Your approach here is valid.

waxlamp (approved):

PR #102, opened 2022-03-09
Remove timeout zoom operation

waxlamp:
This PR removes the programmatic zoom action that is triggered 2.5 seconds after loading an investigation.

### DIFF
--- a/web/src/views/InvestigationDetail.vue
+++ b/web/src/views/InvestigationDetail.vue
@@ -124,7 +124,6 @@ export default defineComponent({
       await store.dispatch.fetchCurrentInvestigation(props.investigation);
       selectedDataset.value = store.state.activeDataset;
       loaded.value = true;
-      setTimeout(() => zoom(6), 2500);
     });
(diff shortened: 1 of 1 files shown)


web/src/views/InvestigationDetail.vue
```diff
@@ -124,7 +124,6 @@ export default defineComponent({
       await store.dispatch.fetchCurrentInvestigation(props.investigation);
       selectedDataset.value = store.state.activeDataset;
       loaded.value = true;
-      setTimeout(() => zoom(6), 2500);
```
naglepuff, line 127: Since we're no longer using this `zoom` function. It can also be removed from line 106.

naglepuff (changes requested):
This change is also included in #103. Doesn't necessarily matter to me if we push this through while that PR is under review.

waxlamp:
Thanks for the review. Since #103 is fixing this too, I'll just close in favor of that.

Issue #92, opened 2022-02-24
Support "text" datasets for annotating an investigation


waxlamp:
_Originally posted by @annehaley in https://github.com/atlascope/atlascope/issues/87#issuecomment-1048103894_:
> A next step would be adding functionality to the frontend to allow the user to place a pin arbitrarily and write a new Dataset with text content to act as a note.

The idea would be to have a dataset type of "note", and include a "text" property in the metadata. A pin involving a note dataset could simply display the associated text on hover, etc.

annehaley:
The Pin model has a `note` text field and the `child` dataset is nullable. Would it suffice to use the Pin like this (no child dataset, just note content) instead of creating a Dataset object just to contain the note? Or is there some use case to do something Dataset-ish with the note?

waxlamp:
> The Pin model has a `note` text field and the `child` dataset is nullable. Would it suffice to use the Pin like this (no child dataset, just note content) instead of creating a Dataset object just to contain the note? Or is there some use case to do something Dataset-ish with the note?

Yeah, the `Pin`'s `note` field seems sufficient. If we didn't do this, we'd have to make the child one of these "note" datasets, right? What's your current vision for how you render the pin's `note` field in-app?

annehaley:
I was thinking pins show up as dots on the context map, and on hover, we can display the content of the `note` field, as well as the info about the `child` dataset if there is one.

waxlamp:
This sounds like a plausible idea. @naglepuff, you've been working on pin display; can you weigh in?

naglepuff:
The "hover to discover" functionality is what I plan to work on next with regards to pin display. I think its a good feature to have, I just need to figure out the "vue" way to do it.

waxlamp:
The "popover mode" of [v-menu](https://vuetifyjs.com/en/components/menus/#popover-menu) is one idea I've used in the past for this sort of thing.

naglepuff:
Closing via #103

Issue #71, opened 2022-02-02
Add Pins to InvestigationDetail


naglepuff:
The investigation detail view currently shows some very basic information about an investigation, including the ability to display any `Dataset` of type `tile_source`.

This view should also display pins for an investigation. Some ideas:
- Display `Pins` in the sidebar. If pins are associated with a derived `Dataset` we could also display pertinent information/visualizations related to that `Dataset`
- Display `Pins` on the `tile_source` `Dataset` for an investigation. This can be done with a feature layer on the `geojs` map, and we could use `geojs` events to drive workflows.  (dependent on resolution of #18)

waxlamp:
Both good ideas. With the current work in flight for the pin model, I think we're on the cusp of being able to experiment with these ideas. @naglepuff I'm assingning this to you, but let me know if you'd prefer someone else to take it.

naglepuff:
@waxlamp I'm happy to take this on.

waxlamp:
Closed by #103.

PR #103, opened 2022-03-09
Show pins as list and display on map

naglepuff:
This PR adds some basic functionality to pins. To demonstrate the UI changes, three pin objects are created with the `populate` command. These pins just contain notes and do not point to child dataset objects.

To test these changes, run the populate script to ensure you have these pins. In front-end application, navigate to the `InvestigationDetail` view for the "Whole Team Investigation." In the sidebar you should see a `Pins` tab, which lists all pins for the investigation, including the three pins added to the populate script as part of these changes. Click the checkbox next to any of these pins to toggle a geojs feature for that pin. It should display with the pin's color. This can be checked from the admin console.

Hover over the pin on the geojs map. The note should appear at your mouse. Moving the mouse outside of the pin will cause the note to disappear again.

Test selecting/deselecting pins in the sidebar. Test collapsing/un-collapsing the sidebar. For all of these tasks, verify that the selected pins in the sidebar reflect the pins shown on the map.

These changes relate to #71 and #92, but does not fully address those issues.

### DIFF
--- a/web/src/views/InvestigationDetail.vue
+++ b/web/src/views/InvestigationDetail.vue
@@ -82,9 +82,10 @@ import {
   ref, defineComponent, onMounted, PropType, computed, watch, Ref,
 } from '@vue/composition-api';
 import useGeoJS from '../utilities/useGeoJS';
+import { Point, postGisToPoint } from '../utilities/utiltyFunctions';
 import store from '../store';
 import InvestigationSidebar from '../components/InvestigationSidebar.vue';
-import { Dataset } from '../generatedTypes/AtlascopeTypes';
+import { Dataset, Pin } from '../generatedTypes/AtlascopeTypes';
 
 export default defineComponent({
   name: 'InvestigationDetail',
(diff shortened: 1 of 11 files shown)


web/src/views/InvestigationDetail.vue
```diff
@@ -151,17 +172,82 @@ export default defineComponent({
       geojsParams.layer.crossDomain = 'use-credentials';
       createMap(geojsParams.map);
       createLayer('osm', geojsParams.layer);
+      clampBoundsX(false);
+    });
+    // #endregion datasets
+
+    // #region pin
```
waxlamp: Is this like a folding support for an IDE or something?

web/src/views/InvestigationDetail.vue
```diff
@@ -151,17 +172,82 @@ export default defineComponent({
...
+    /* eslint-enable */
+    watch(selectedPins, (pinList) => {
+      if (!featureLayer) {
+        featureLayer = createLayer('feature', { features: ['point', 'line', 'polygon'] });
+      }
+      const pinFeatureData = pinList.map((pin) => {
+        let pinLocation: Point | undefined = postGisToPoint(pin.child_location);
+        if (!pinLocation) {
+          pinLocation = { x: 0, y: 0 };
+        }
```
waxlamp: I think the following should work (but I haven't tested it):

```suggestion
        const pinLocation: Point = postGisToPoint(pin.child_location) || { x: 0, y: 0 };
```

web/src/views/InvestigationDetail.vue
```diff
@@ -151,17 +172,82 @@ export default defineComponent({
...
+          note: pin.note,
+        };
+      });
+      if (!pinFeature) {
+        /* eslint-disable */
+        pinFeature = featureLayer.createFeature('point')
+          .data(pinFeatureData)
+          .position((pin: any) => ({ x: pin.x, y: pin.y }))
+          .style({
+            // reasonable default?
```
waxlamp: Yes, this is fine.

```suggestion
```

waxlamp (approved):
Good stuff. Lots here to build off of!

web/src/views/InvestigationDetail.vue
naglepuff: Yeah, it is. I was not thinking of the fact that there's no standard tool people use. I don't mind removing the region markers if that's best practice.

waxlamp: It's probably best to remove them.

What IDE are you using? I know vim does this sort of thing in-band but in modern editors you ought to be able to do it out-of-band (so to speak).

naglepuff: vscode

naglepuff: Removed in e8f054f

mcovalt (approved):

PR #104, opened 2022-03-09
Pin minio to a version that doesn't break makemigrations

waxlamp:


marySalvi (approved):

annehaley (approved):
We did the same thing on MIQA

docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
   minio:
     logging:
       driver: none
-    image: bitnami/minio:2022.3.3
+    image: minio/minio:RELEASE.2022-03-03T21-21-16Z
```
annehaley, line 28: Is there not a x.x.x tag for this release? If not, is this even a formal release that we should pin to?

waxlamp: For better or worse, all of the [minio image release tags](https://github.com/minio/minio/tags) look like this. These are, AFAICT, the formal releases.

annehaley: Ok, thanks for clarifying. This looks good then, consider my approval refreshed :grin:

marySalvi (approved):

PR #105, opened 2022-03-10
Improve subimage ergonomics

waxlamp:
This PR makes the following attempted improvements to the subimage infrastructure:
- move the subimage creation machinery to the `Dataset` model (and access it therefrom in the associated endpoint)
- make the endpoint operate on a specific `Dataset`

This has two basic effects:
1. It allows for easier creation of subimage datasets in the populate script.
2. It simplifies the serializer used to create subimages.

@marySalvi, please let me know if this is a valid improvement on the subimage machinery.
@annehaley, please advise on whether this is good Django practice.

~(This PR depends on #99; merging that PR will make the diff for this one much easier to digest.)~ #99 has been merged.

### DIFF
--- a/atlascope/core/models/dataset.py
+++ b/atlascope/core/models/dataset.py
@@ -114,9 +114,7 @@ class DatasetSubImageSerializer(serializers.Serializer):
     y1 = serializers.IntegerField(required=True)
 
     def validate_original_dataset_id(self, value):
-        if not Dataset.objects.filter(id=value).exists():
-            raise ValidationError(f'{value} does not exist. Must use existing dataset')
-        return str(value)
+        return Dataset.objects.get(id=value)
 
 
 @admin.register(Dataset)
--- a/atlascope/core/models/dataset.py
+++ b/atlascope/core/models/dataset.py
@@ -43,6 +43,24 @@ class Dataset(TimeStampedModel, models.Model):
         if not self.name:
             self.name = importer_obj.dataset_name or f'{importer} {self.id}'
 
+    def subimage(self, x0: int, x1: int, y0: int, y1: int) -> 'Dataset':
+        metadata = {
+            'x0': x0,
+            'x1': x1,
+            'y0': y0,
+            'y1': y1,
+        }
+
+        dataset = Dataset(
+            name=f'{self.name} Subimage ({x0}, {y0}) -> ({x1}, {y1})',
+            metadata=metadata,
+            source_dataset=self,
+            content=self.content,
+            dataset_type="subimage"
+        )
+
+        return dataset
+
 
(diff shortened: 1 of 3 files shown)


atlascope/core/models/dataset.py
```diff
@@ -43,6 +43,24 @@ def perform_import(self, importer="UploadImporter", **kwargs):
...
+            'x1': x1,
+            'y0': y0,
+            'y1': y1,
+        }
+
+        dataset = Dataset(
+            name=f'{self.name} Subimage ({x0}, {y0}) -> ({x1}, {y1})',
+            metadata=metadata,
+            source_dataset=self,
+            content=self.content,
```
annehaley, line 58: Are we actually going to crop the content eventually or just read the bounds from the metadata?

atlascope/core/rest/endpoints/dataset_endpoints.py
```diff
@@ -39,19 +39,13 @@ def create(self, request):
...
-        new_dataset_obj = Dataset(
-            name=f'{original_dataset.name} Subimage',
-            metadata=meta,
-            source_dataset=original_dataset,
-            dataset_type="subimage",
-        )
-        new_dataset_obj.save()
 
-        return Response(DatasetSerializer(new_dataset_obj).data, status=status.HTTP_201_CREATED)
+        original = Dataset.objects.get(id=pk)
```
annehaley: You can just call `self.get_object()`

atlascope/core/rest/endpoints/dataset_endpoints.py
```diff
@@ -39,19 +39,13 @@ def create(self, request):
         return Response(DatasetSerializer(new_dataset_obj).data, status=status.HTTP_201_CREATED)
 
     @swagger_auto_schema(request_body=DatasetSubImageSerializer())
-    @action(detail=False, methods=['POST'])
-    def subimage(self, request):
+    @action(detail=True, methods=['POST'])
```
annehaley, line 42: I do think this is better API design. I like this refactoring.

annehaley (approved):

atlascope/core/models/dataset.py
waxlamp: For now I just wanted to be able to read bounds from metadata--that's the simplest path to be able to work with subimages. As soon as that becomes painful for whatever reason, we can have a method to "reify" these subimages into full-fledged images.

atlascope/core/rest/endpoints/dataset_endpoints.py
waxlamp: Oh, neat! See b6fa31ab15767e3e801c8eea27d4e9a1e6cca257.

marySalvi (approved):

PR #107, opened 2022-03-14
Add test tile server for colors

mcovalt:
This PR adds an extra tile server just for testing. It is hosted on port 9005.

- `/{color}.*` returns a _solid_ color tile. This is useful for spotting the small nested color that is difficult to locate in the speckled case.
- `/speckled-{color}.*` returns a _speckled_ color tile. This is useful to ensure perspectives are not warped and identifying when zooming retrieves new tiles compared to using interpolated scaling.

`color` is one of "red", "blue", "green", or "yellow"


waxlamp (approved):
Can this be shimmed into the Dataset architecture? Or does it have to be a special case usage of these particular tile service endpoints?

mcovalt:
> Can this be shimmed into the Dataset architecture? Or does it have to be a special case usage of these particular tile service endpoints?

This service doesn't represent a real dataset because a tile is always returned regardless of the path. This allows representing an infinitely large and zoomable image. Our current endpoints only allow viewing image files. These will always have a finite dimension and number of tiles. This is a separate service because of this difference.

waxlamp:
I guess I was thinking more like: can we rig a `Dataset` with a specific ID (or name, or whatever) that triggers a special path in the tile server logic that "shunts" those datasets to this endpoint? If so, we can treat these data sources like any other (creating an `Investigation` that features them, for example).

mcovalt:
We could incorporate it in some way. It would also require implementing a response for our tile metadata endpoint. This endpoint returns things like dimensions, number of zoom levels, etc. We'd need to consider what these data would look like.

naglepuff:
What if we just made some big tiffs with the colored tiles and added those to the populate script as datasets?

mcovalt:
> What if we just made some big tiffs with the colored tiles and added those to the populate script as datasets?

That's the easiest way to do this, I think.

waxlamp:
That wouldn't get us the nifty infinite zoom feature the tile servers give us, but if it gets us most of the way there for testing and demo purposes, I think it's a good bang for the effort.

PR #109, opened 2022-03-14
Modify tile endpoints to accept `channels` and `colors` querystring args

annehaley:
This PR uses `fsspec` to locally cache the remote `dataset.content.url` and replaces the `GDALFileTileSource` with `OMETiffFileTileSource`. It may be necessary to use `if` logic and support both.

Supplying the `channel` kwarg to the tile endpoint will return the tile from only the channel corresponding to that index. This will enable us to color and composite the channels on the frontend rather than the backend. The metadata endpoint supplies `metadata.frames`, which describes the channels as they are defined within the TIFF.

### DIFF
--- a/atlascope/core/rest/endpoints/tile_endpoints.py
+++ b/atlascope/core/rest/endpoints/tile_endpoints.py
@@ -112,9 +112,11 @@ class TileView(GenericAPIView, mixins.RetrieveModelMixin):
                 channels = channels.split(',')
             else:
                 channels = range(len(tile_source.getMetadata()['frames']))
-            colors = self.request.query_params.get('colors').split(',') or range(
-                self.default_colors
-            )
+            colors = self.request.query_params.get('colors')
+            if colors:
+                colors = [f'#{color}' for color in colors.split(',')]
+            else:
+                colors = self.default_colors
             composite = None
             for channel, color in list(zip(channels, colors)):
                 tile = tile_source.getTile(
@@ -126,7 +128,7 @@ class TileView(GenericAPIView, mixins.RetrieveModelMixin):
                 tile_data = numpy.array(PIL.Image.open(io.BytesIO(tile)))
                 if not composite:
                     composite = PIL.Image.new('RGBA', tile_data.shape, '#000000')
-                mask_color = PIL.Image.new('RGBA', tile_data.shape, f'#{color}')
+                mask_color = PIL.Image.new('RGBA', tile_data.shape, color)
                 mask = PIL.Image.fromarray(tile_data)
                 composite = PIL.Image.composite(mask_color, composite, mask)
             buf = io.BytesIO()
--- a/atlascope/core/rest/endpoints/tile_endpoints.py
+++ b/atlascope/core/rest/endpoints/tile_endpoints.py
@@ -1,8 +1,10 @@
+import fsspec
+
 from django.urls import path
 from drf_yasg import openapi
 from drf_yasg.utils import swagger_auto_schema
 from large_image.exceptions import TileSourceError
-from large_image_source_gdal import GDALFileTileSource
+from large_image_source_ometiff import OMETiffFileTileSource
 from rest_framework import mixins
 from rest_framework.exceptions import APIException, NotFound
 from rest_framework.generics import GenericAPIView
@@ -19,7 +21,11 @@ class TileMetadataView(GenericAPIView, mixins.RetrieveModelMixin):
 
     def get(self, *args, **kwargs):
         dataset = self.get_object()
-        tile_source = GDALFileTileSource(f'/vsicurl/use_head=no&url={dataset.content.url}')
+        cached = fsspec.open_local(
+            f'simplecache::{dataset.content.url}',
+            filecache={'cache_storage': '/tmp/files'},
+        )
+        tile_source = OMETiffFileTileSource(cached[0])
         serializer = self.get_serializer(tile_source)
         return Response(serializer.data)
 
@@ -73,11 +79,13 @@ class TileView(GenericAPIView, mixins.RetrieveModelMixin):
     )
     def get(self, *args, x=None, y=None, z=None, **kwargs):
         dataset = self.get_object()
-        tile_source = GDALFileTileSource(
-            f'/vsicurl/use_head=no&url={dataset.content.url}', encoding='PNG'
+        cached = fsspec.open_local(
+            f'simplecache::{dataset.content.url}',
+            filecache={'cache_storage': '/tmp/files'},
         )
+        tile_source = OMETiffFileTileSource(cached[0])
         try:
-            tile = tile_source.getTile(x, y, z)
+            tile = tile_source.getTile(x, y, z, frame=kwargs.get('channel'))
         except TileSourceError as e:
             error_msg = str(e)
             for missing_msg in (
--- a/atlascope/core/rest/endpoints/tile_endpoints.py
+++ b/atlascope/core/rest/endpoints/tile_endpoints.py
@@ -1,8 +1,7 @@
-import fsspec
-
 from django.urls import path
 from drf_yasg import openapi
 from drf_yasg.utils import swagger_auto_schema
+import fsspec
 from large_image.exceptions import TileSourceError
 from large_image_source_ometiff import OMETiffFileTileSource
 from rest_framework import mixins
--- a/atlascope/core/rest/endpoints/tile_endpoints.py
+++ b/atlascope/core/rest/endpoints/tile_endpoints.py
@@ -102,7 +102,7 @@ class TileView(GenericAPIView, mixins.RetrieveModelMixin):
         channels = self.request.query_params.get('channels').split(',') or range(
             len(tile_source.getMetadata()['frames'])
         )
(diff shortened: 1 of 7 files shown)


annehaley:
Current CI failures are due to a missing dependency (local runs of tox tests succeed, though). 
```
ImportError: Failed to find TIFF header file (may be need to run: sudo apt-get install libtiff4-dev)
```
I tried adding `libtiff4-dev`, `libtiff-dev`, `libtiff`, and `libtiff5-dev`. Any suggestions on installing this correctly?

mcovalt:
`libtiff-dev` is probably what we want. [It's `libtiff` version 4.1.0 in Ubuntu 20.04](https://packages.ubuntu.com/focal/libtiff-dev) which is [supported by pylibtiff](https://github.com/pearu/pylibtiff/blob/a9769ec5d9533b14915d65b3e776dd95564b8d11/libtiff/tiff_h_4_1_0.py). It'll need to be installed in both the [CI](https://github.com/atlascope/atlascope/blob/4cc0494ec1cea8953cd4c19c33d517c9c5b9083f/.github/workflows/ci.yml) and in the [Dockerfile](https://github.com/atlascope/atlascope/blob/4cc0494ec1cea8953cd4c19c33d517c9c5b9083f/dev/django.Dockerfile).

mcovalt:
`libtiff-dev` is probably what we want. [It's `libtiff` version 4.1.0 in Ubuntu 20.04](https://packages.ubuntu.com/focal/libtiff-dev) which is [supported by pylibtiff](https://github.com/pearu/pylibtiff/blob/a9769ec5d9533b14915d65b3e776dd95564b8d11/libtiff/tiff_h_4_1_0.py). It'll need to be installed in both the [CI](https://github.com/atlascope/atlascope/blob/4cc0494ec1cea8953cd4c19c33d517c9c5b9083f/.github/workflows/ci.yml) and in the [Dockerfile](https://github.com/atlascope/atlascope/blob/4cc0494ec1cea8953cd4c19c33d517c9c5b9083f/dev/django.Dockerfile).

atlascope/core/rest/endpoints/tile_endpoints.py
```diff
@@ -19,7 +20,11 @@ class TileMetadataView(GenericAPIView, mixins.RetrieveModelMixin):
 
     def get(self, *args, **kwargs):
         dataset = self.get_object()
-        tile_source = GDALFileTileSource(f'/vsicurl/use_head=no&url={dataset.content.url}')
+        cached = fsspec.open_local(
+            f'simplecache::{dataset.content.url}',
+            filecache={'cache_storage': '/tmp/files'},
+        )
+        tile_source = OMETiffFileTileSource(cached[0])
```
mcovalt: We'll likely receive 10+ separate requests instantaneously for a dataset. Will all 10+ of those requests download a file? Or is there a lock so that only 1 does?

annehaley: Each request would download a file with one channel. That's the crux of the tradeoff between compositing the channels client side versus server side. Which do you think is better for our use case?

mcovalt: Sorry, I should've been more clear. I'm trying to understand what happens in the `fsspec.open_local` command. I believe I understand the basics of how it works:

- A file is created with a name based on the `dataset.content.url` if it doesn't exist. (cache-miss)
- That file is placed in `/tmp/files`.
- The remote content at `dataset.content.url` is downloaded and written to that file.
- The file is not created and the download does not occur if the file already exists. (cache-hit)

The idea is that the file will be downloaded only once and then subsequent requests can just read from that cached download. I'm considering the typical scenario for a cache-miss in this case:

- The user visits a page with GeoJS for a new dataset.
- GeoJS then fills up its canvas by requesting tons of tiles for that dataset. These requests are done in parallel:
  - `http://atlascope.com/dataset/some-uuid/tiles/0/0/0.png`
  - `http://atlascope.com/dataset/some-uuid/tiles/1/0/0.png`
  - `http://atlascope.com/dataset/some-uuid/tiles/1/1/0.png`
  - `http://atlascope.com/dataset/some-uuid/tiles/1/0/1.png`
  - ...
- The Django server handles each of these requests in parallel; one request per thread.

I'm guessing there will be a cache-miss for each of these requests the first time it is downloaded.
1. Does each thread of the Django server begin downloading the dataset?
2. Does one thread "win", create a file, and the other threads idle while waiting for that file to download? (file lock)

I think in both cases we'll exhaust all request handling threads and won't be able to serve tiles or web pages until the download is done. In the first case I think we'll just do that "worse". I don't know if these are show stoppers for a demo, but I think we ought to know exactly what is going on if it fails during a demo.

> Each request would download a file with one channel. That's the crux of the tradeoff between compositing the channels client side versus server side. Which do you think is better for our use case?

For this question, I can't think of a reason one or the other would be preferred. Whatever is easiest makes the most sense to me.

annehaley: Oh I see. Yes, your description is in line with what I read from the docs, but I'm not sure how it handles races between parallel requests. It's something that we can come back to, like you say. 

Let's explore what it would look like to do client-side compositing first and fall back to server-side compositing, since I've tried that and I know how to implement it. I think it would be nice to allow the user to toggle channels and change colors without making subsequent requests.

mcovalt (approved):
I added ad593064e8740b92fe660d42bc087a62dd5ac3bc to address some packaging issues:

- https://github.com/girder/large_image/pull/799
- https://github.com/girder/large_image_wheels/issues/29

Take a look at ad593064e8740b92fe660d42bc087a62dd5ac3bc before merging.

naglepuff (approved):

naglepuff:
@annehaley If I make a request for a tile with no query string parameters, I get the following error:
```
Internal Server Error: /api/v1/datasets/ed0fea81- log.py:224
django_1     |                     5c61-4c0d-a9aa-570481ea0059/tiles/1/1/1.png                 
django_1     |                     Traceback (most recent call last):                          
django_1     |                       File "/usr/local/lib/python3.8/dist-packages/dj           
django_1     |                     ango/core/handlers/exception.py", line 47, in               
django_1     |                     inner                                                       
django_1     |                         response = get_response(request)                        
django_1     |                       File "/usr/local/lib/python3.8/dist-packages/dj           
django_1     |                     ango/core/handlers/base.py", line 181, in                   
django_1     |                     _get_response                                               
django_1     |                         response = wrapped_callback(request,                    
django_1     |                     *callback_args, **callback_kwargs)                          
django_1     |                       File "/usr/local/lib/python3.8/dist-packages/dj           
django_1     |                     ango/views/decorators/csrf.py", line 54, in                 
django_1     |                     wrapped_view                                                
django_1     |                         return view_func(*args, **kwargs)                       
django_1     |                       File "/usr/local/lib/python3.8/dist-packages/dj           
django_1     |                     ango/views/generic/base.py", line 70, in view               
django_1     |                         return self.dispatch(request, *args,                    
django_1     |                     **kwargs)                                                   
django_1     |                       File "/usr/local/lib/python3.8/dist-packages/re           
django_1     |                     st_framework/views.py", line 509, in dispatch               
django_1     |                         response = self.handle_exception(exc)                   
django_1     |                       File "/usr/local/lib/python3.8/dist-packages/re           
django_1     |                     st_framework/views.py", line 469, in                        
django_1     |                     handle_exception                                            
django_1     |                         self.raise_uncaught_exception(exc)                      
django_1     |                       File "/usr/local/lib/python3.8/dist-packages/re           
django_1     |                     st_framework/views.py", line 480, in                        
django_1     |                     raise_uncaught_exception                                    
django_1     |                         raise exc                                               
django_1     |                       File "/usr/local/lib/python3.8/dist-packages/re           
django_1     |                     st_framework/views.py", line 506, in dispatch               
django_1     |                         response = handler(request, *args, **kwargs)            
django_1     |                       File "/opt/django-project/atlascope/core/rest/e           
django_1     |                     ndpoints/tile_endpoints.py", line 102, in get               
django_1     |                         channels = self.request.query_params.get('cha           
django_1     |                     nnels').split(',') or range(                                
django_1     |                     AttributeError: 'NoneType' object has no                    
django_1     |                     attribute 'split'            
```

annehaley:
> @annehaley If I make a request for a tile with no query string parameters, I get the following error:

Addressed in [78220af](https://github.com/atlascope/atlascope/pull/109/commits/78220afc67e042a319a9b9594811a15f19834a41)

atlascope/core/rest/endpoints/tile_endpoints.py
```diff
@@ -19,7 +25,14 @@ class TileMetadataView(GenericAPIView, mixins.RetrieveModelMixin):
...
         dataset = self.get_object()
-        tile_source = GDALFileTileSource(f'/vsicurl/use_head=no&url={dataset.content.url}')
+        cached = fsspec.open_local(
+            f'simplecache::{dataset.content.url}',
+            filecache={'cache_storage': '/tmp/files'},
+        )
+        try:
+            tile_source = OMETiffFileTileSource(cached[0])
+        except TileSourceError:
+            tile_source = TiffFileTileSource(cached[0])
```
annehaley, line 35: This is the implementation of simple file caching with `fsspec`

Issue #110, opened 2022-03-18
Multiple relationships between Investigation and Dataset


naglepuff:
With our current data model, `Datasets` can be related to an `Investigation` in a couple of ways.

1. Via the `Investigation.datasets` property
2. Via a `DatasetEmbedding` as either a `DatasetEmbedding.parent` or `DatasetEmbedding.child`

I imagine that eventually we want to ensure that for all of an Investigation's embeddings, the `parent` and `child` datasets are also included in (1).

waxlamp:
In our meeting this week, we discussed reinterpreting `Investigation.datasets` as the list of all "root" datasets for an investigation. The main way this interpretation will show up in Atlascope is by displaying all the root datasets side by side in the Investigation detail view, and by using these roots as the basis on which to create the dataset embedding trees.

PR #111, opened 2022-03-23
Fix pin-related bugs

naglepuff:
This PR fixes some bugs that I've observed with the pin list and pin display in the Investigation detail view.

**Pin Selection**
Pin selection is now cleared when the active dataset changes via the drop down. Prior to this change, pins would remain selected, as indicated by the checkbox, even after the dataset being displayed changed. 

**Pin Display**
Feature layer creation is now more consistent. Previously, if the active dataset was changed, the feature layer would remain, but the map it was associated with would be destroyed. The layer would not be associated with the new map that was created as a result of the dataset change. Now, the feature layer is destroyed when the active dataset changes, and a new one is created to show pins on the new dataset.

**Pins Respect Active Dataset**
Now selected pins are filtered based on active dataset before being displayed on the dataset. This works fine for now, but once the embeddings view is merged into the current interation of the `InvestigationDetail` view, we will need a more sophisticated filter (and likely a more sophisticated way to determine location).

### DIFF
--- a/web/src/views/InvestigationDetail.vue
+++ b/web/src/views/InvestigationDetail.vue
@@ -141,14 +141,24 @@ export default defineComponent({
     const selectedDataset: Ref<Dataset | null> = ref(null);
     const activeDataset = computed(() => store.state.activeDataset);
     const tilesourceDatasets = computed(() => store.getters.tilesourceDatasets);
+    /* eslint-disable */
+    let featureLayer: any;
+    let pinFeature: any;
+    /* eslint-enable */
 
     function activeDatasetChanged(newActiveDataset: Dataset) {
       selectedDataset.value = newActiveDataset;
       store.dispatch.setActiveDataset(newActiveDataset);
     }
 
-    function drawMap(dataset: Dataset | null) {
+    function tearDownMap() {
       exit();
+      featureLayer = null;
+      pinFeature = null;
+    }
+
+    function drawMap(dataset: Dataset | null) {
+      tearDownMap();
       if (!dataset || !dataset.id) {
         return;
       }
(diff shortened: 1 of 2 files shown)


web/src/views/InvestigationDetail.vue
```diff
@@ -189,19 +190,30 @@ export default defineComponent({
...
-      }
-      const pinFeatureData = pinList.map((pin) => {
-        const pinLocation: Point = postGisToPoint(pin.child_location) || { x: 0, y: 0 };
+    function getPinsToDisplay() {
+      // TODO: as we move towards embedding multiple datasets into the view,
+      // we will need a more sophisticated way to determine which pins to render
+      // and determining where they should be rendered
+      const selectedPinsForActiveDataset = selectedPins.value.filter(
+        (pin) => pin.parent === activeDataset.value?.id,
+      );
```
waxlamp, line 199: How would you end up selecting pins that aren't part of the active dataset? Aren't all of the pins in an investigation part of a dataset that would be active? Or is that something that might be true after our "reinterpretation" of `Investigation.datasets` as per #110?

waxlamp (approved):
Looks good. Thanks!

web/src/views/InvestigationDetail.vue
naglepuff: The pins that show up in the sidebar are related to the current investigation. `activeDataset` here is the dataset that is currently displayed by `geojs`. It is possible that pins for an investigation have different parent datasets. Since we can only show one dataset at a time (at least in the `InvestigationDetail`) right now, its possible to select pins that aren't related to the currently visible dataset.

The original change prevented such selected pins from showing up on the map, but perhaps a better change is 072af39, which disabled selection for pins whose parent dataset is not the active dataset.

Once we implement our new interpretation of `Investigation.datasets`, we'll need to update how we determine which pins are eligible to be selected, and how those pins are displayed on the map. (might be with a feature layer per tile layer, or doing math to place the pins on embedded datasets)

PR #113, opened 2022-03-24
Investigation tour

marySalvi:
This still needs a bit of work but its a first pass at the tour concept


atlascope/core/models/tour.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class TourSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Tour
+        fields = '__all__'
+
+
+class TourCreateSerializer(serializers.Serializer):
+    way_points = serializers.ListField(child=serializers.ListField(child=serializers.IntegerField()))
```
marySalvi: This specifically needs to be reworked because the IntegerField cannot be used here.

annehaley:
Since this is supposed to be a conversation starter, could you explain the intent for the `way_points` field? What kind of information could it contain?

atlascope/core/models/tour.py
```diff
@@ -0,0 +1,33 @@
...
+        on_delete=models.CASCADE,
+        related_name='tour',
+    )
+    way_points = ArrayField(
+        ArrayField(
+            gisModels.PointField(null=True),
+            null=True
+        ),
+        null=True
+    )
```
waxlamp: Why a 2D array of points?

Two thoughts:
1. Ideally we'd create a `Waypoint` class (with a `Point` and an integer zoom level) that could be the field type, though I don't really know the right way to do that sort of thing.
2. It looks like `PointField` supports 3D points by default; we could certainly abuse that to represent a 2D location plus a zoom level.

atlascope/core/models/tour.py
waxlamp: Shouldn't it be a PointField serializer, instead of IntegerField?

waxlamp (commented):
Once we figure out the right representation for a waypoint, a lot of this should flow easier. At a minimum, we need to capture a location and zoom level; later on we might want to include a note or some other free form information as well.

marySalvi:
@annehaley @waxlamp 
This is far from complete (not actually accessible from an endpoint yet), but I would like your input regarding the refactor/restructuring of tour and waypoint 951df33d7bc28ee3d7df943f2cc2c4e65de25e1d

atlascope/core/models/tour.py
```diff
@@ -0,0 +1,16 @@
+# from typing import List
+from uuid import uuid4
+
+from django.db import models
+from rest_framework import serializers
+
+
+class Tour(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
```
waxlamp: If #120 is approved, you'll want to strike this line.

atlascope/core/models/tour.py
```diff
@@ -0,0 +1,16 @@
...
+    way_points = models.ManyToManyField('Waypoint', through='Investigation')
```
waxlamp: I don't think this is right--we don't want to use `Investigation` as the through-table. I'm thinking we want to invent a new model to handle the relationship between several `Waypoint`s and a `Tour`. Check out the [membership example](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.ManyToManyField.through) in the Django docs. Let's discuss in person if you want to dig into this.

atlascope/core/models/investigation.py
```diff
@@ -12,6 +12,8 @@ class Investigation(TimeStampedModel, models.Model):
     description = models.TextField(max_length=5000, blank=True)
     datasets = models.ManyToManyField('Dataset', related_name='investigations')
     notes = models.TextField(max_length=5000, blank=True)
+    way_point = models.ForeignKey('Waypoint', on_delete=models.CASCADE, null=True)
+    tour = models.ForeignKey('Tour', on_delete=models.CASCADE, null=True)
```
waxlamp: This confused me until I saw your through table setup below--see that comment for my thoughts on this change.

atlascope/core/models/waypoint.py
```diff
@@ -0,0 +1,10 @@
+from uuid import uuid4
+
+from django.contrib.gis.db import models as gisModels
+from django.db import models
+
+
+class Waypoint(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
```
waxlamp: Same comment as above, re #120 and striking this line.

waxlamp (commented):
Good start. The main sticking point is my intuition that we need to create a custom through model to mediate between `Tour` and `Waypoint`.

atlascope/core/models/tour.py
marySalvi: resolved in 2d813ecddf8363d972acfdf8626c85516203de80

atlascope/core/models/tour.py
marySalvi: resolved with 56d7e61af852626496a99eb9a5c4ec4b7b136f08

atlascope/core/models/tour.py
marySalvi: resolved with 56d7e61af852626496a99eb9a5c4ec4b7b136f08

atlascope/core/models/investigation.py
marySalvi: resolved with 56d7e61af852626496a99eb9a5c4ec4b7b136f08

atlascope/core/models/tour.py
marySalvi: resolved with a29845873608f22d5ba2d213129b54560c2522ae

atlascope/core/models/waypoint.py
marySalvi: resolved with a29845873608f22d5ba2d213129b54560c2522ae

atlascope/core/rest/endpoints/investigation_endpoints.py
```diff
@@ -38,3 +38,18 @@ def jobs(self, request, pk=None):
     def embeddings(self, request, pk=None):
         payload = DatasetEmbeddingSerializer(self.get_object().embeddings.all(), many=True).data
         return Response(payload, status=status.HTTP_200_OK)
+
+    # @swagger_auto_schema(request_body=TourCreateSerializer())
+    # @action(detail=True, methods=['POST'])
+    # def tour(self, request, pk=None):
+    #     serializer = TourCreateSerializer(data=request.data)
```
annehaley: Why did you decide to comment this out?

atlascope/core/models/waypoint.py
```diff
@@ -0,0 +1,11 @@
+from django.contrib import admin
+from django.contrib.gis.db import models as gis_models
+from django.db import models
+
+
+class Waypoint(models.Model):
+    location = gis_models.PointField(null=True)
+    zoom = models.IntegerField(null=True)
+
```
annehaley, line 9: Wouldn't we want a text/char field on each waypoint, too? At each zoomed location, there's some bit of important information that should be shared in the tour and displayed when that waypoint is active.

atlascope/core/models/tour.py
```diff
@@ -0,0 +1,22 @@
...
+    )
+
+
+admin.site.register(Tour)
+
+
+class TourSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Tour
+        fields = '__all__'
```
annehaley: I think you'll need to a Waypoint serializer that you invoke here so that the waypoint information is nested within the serialized tour object. Otherwise you'll just get a list of integers representing each waypoint.

annehaley (changes requested):

atlascope/core/models/waypoint.py
waxlamp: I don't think we know yet what (if anything) we'll want to be displayed during an executed tour, so the best thing right now is to keep the code as simple as possible. We will revisit this when we see what it's like to "run" a tour, and at that point we may add more fields here.

atlascope/core/rest/endpoints/investigation_endpoints.py
```diff
@@ -4,7 +4,7 @@
 from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet
 
-from atlascope.core.models import (
+from atlascope.core.models import (  # Tour,; TourCreateSerializer,; TourSerializer,
```
waxlamp: Can we get rid of this comment, or does it have some effect on the tooling or something?

atlascope/core/rest/endpoints/investigation_endpoints.py
waxlamp: If this stuff isn't needed for the current PR, let's cut it out.

atlascope/core/models/tour.py
waxlamp: Sounds like a good idea. @marySalvi, I'm happy to look at this with you; I believe this is the relevant documentation: https://www.django-rest-framework.org/api-guide/relations/#nested-relationships.

atlascope/core/management/populate/tours.json
```diff
@@ -0,0 +1,31 @@
+[{
```
waxlamp, line 1: This is good for now, but we'll want to update this to a meaningful tour as we put together the pieces for our example application workflow.

waxlamp (changes requested):
This looks pretty good overall; I think we should address Anne's comments about the tour serializer before merging.

atlascope/core/rest/endpoints/investigation_endpoints.py
marySalvi: Commented create stuff removed and read endpoint added in 19ba0032d9cea5ca607ae52bea612feec0d07ebd

atlascope/core/models/tour.py
marySalvi: resolved in 99d34aed6212af61cfb082bacedac18ec38537e0

atlascope/core/rest/endpoints/investigation_endpoints.py
marySalvi: removed in 19ba0032d9cea5ca607ae52bea612feec0d07ebd

annehaley (approved):
This looks great!

waxlamp (approved):

PR #114, opened 2022-03-24
Simplify populate script

waxlamp:
This PR simplifies and dumbifies the populate script. This has two main effects:
1. It makes it easier to understand and modify
2. It enables us to insert more customized logic for specific kinds of resources that may show up in the future (e.g., subimages)

Note that I have also simplified some of the (implicit) schema used in the JSON files driving the populate script.

Reviewers: please look for anything I may have done here that will break any assumptions needed for populated data to work properly in the app.

### DIFF
--- a/atlascope/core/management/commands/populate.py
+++ b/atlascope/core/management/commands/populate.py
@@ -114,13 +114,39 @@ def populate_datasets(jsonfile):
     print("")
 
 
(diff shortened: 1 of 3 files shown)


annehaley (approved):
I'm a big fan of this new organization. Separating each model into its own function makes the steps to save more explicit. This script really needed this.

PR #115, opened 2022-03-25
Sort rest results by id everywhere

### LINKED ISSUE #39: Nondeterministic test results
Tests on `main` [were failing](https://github.com/atlascope/atlascope/runs/4781741058) following the merge of #33. Simply re-running the tests without changing the code-base caused tests [to succeed](https://github.com/atlascope/atlascope/runs/4781762885).

annehaley:
Resolves #39 
I found that test failures were coming from instances like this, where two investigations had the same name and were therefore in a different sort order between the api and the expected test results:
image


waxlamp (approved):
Excellent detective work 🕵🏾‍♀️

PR #117, opened 2022-03-25
Add menu to edit visible channels and colors

naglepuff:
This is a basic implementation of a component that allows changing which tiff channels are visible, and what color each channel should be. Currently it works with the `InvestigationDetail` view to allow customizing channels for the current active dataset. 

In that view, click the new `FRAMES` button and explore toggling various  channels and changing their color.


mcovalt (approved):
Looks great!

PR #120, opened 2022-03-31
Use regular, autoincrementing integers for primary keys instead of UUID

waxlamp:
This PR simplifies our models to use Django's default. Using UUID4 values means that every time you run the populate script, you need to start the web app from the top level in order to find your investigations, datasets, etc. This gets to be annoying while working on something that involves changing the populate script, and it makes development simpler in general (by letting us not worry about the `id` field on any model anymore).

Note: because this PR changes historical migrations, to apply this the easiest thing is to run `manage.py reset_db`, run the migrations, and then run the populate script.


atlascope/core/migrations/0002_base_models.py
```diff
@@ -24,8 +22,8 @@ class Migration(migrations.Migration):
             fields=[
                 (
                     'id',
-                    models.UUIDField(
-                        default=uuid.uuid4, editable=False, primary_key=True, serialize=False
+                    models.AutoField(
+                        auto_created=True, primary_key=True, serialize=False, verbose_name='ID'
```
annehaley, line 26: Did you choose to include this verbose name argument or was that generated by makemigrations? I'm concerned that the all-caps version may violate some linting rules if we reference this field that way.

annehaley (approved):

marySalvi (approved):

atlascope/core/migrations/0002_base_models.py
waxlamp: I am pretty sure that came from `makemigrations`. I probably didn't run `makemigrations` myself here, and instead lifted the `AutoField` line from another Girder4 project's 0002 migration. I think I also created a random model in the codebase just to see what popped out of `makemigrations`, and maybe that's where I lifted this from. Sorry, I can't quite remember what I did, but if you're worried about it you can run your own such test to see what it spits out.

annehaley: If you lifted the `AutoField` line from somewhere else, it's probably safer to take out the `verbose_name` argument from these lines.

waxlamp: So, I did the following as a test: in the jobs model file, I created a new model class and added a foreign key field to the Job model pointing to it (so that makemigrations would pick it up). The generated migration has the verbose name bit in it, so I am thinking this is probably fine as is. What do you think?

annehaley: Ok that should be fine, then. I doubt we will use the uppercase version

PR #121, opened 2022-03-31
Revamp populated data

waxlamp:
This PR attempts to restructure our populated data specs to be more reflective of the things we're working on. This mainly involves renaming the datasets to reflect their original sources, and renaming investigations to refer to features. Each investigation is set up with the materials needed to work on that feature (more or less). Comments on the investigations I have here, as well as ones I may be missing, welcome.

TODO:
- [x] add some simple json schemas to make sure the spec files are healthy

Depends on #120.

### DIFF
--- a/atlascope/core/management/populate/datasets.json
+++ b/atlascope/core/management/populate/datasets.json
@@ -7,6 +7,7 @@
     },
     {
         "name": "HTA9_1_BA_F_ROI02",
+        "id": 205,
         "description": "VANDY test dataset 1 (https://styx.neurology.emory.edu/girder/#item/60e84a1868f4fe34fc7eef71)",
         "content": "HTA9_1_BA_F_ROI02.ome.tif"
     },
@@ -18,6 +19,7 @@
     },
     {
         "name": "HTA9_1_BA_F_ROI04",
+        "id": 206,
         "description": "VANDY test dataset 2 (https://styx.neurology.emory.edu/girder/#item/60e84a1968f4fe34fc7eef79)",
         "content": "HTA9_1_BA_F_ROI04.ome.tif"
     },
--- a/atlascope/core/management/populate/datasets.json
+++ b/atlascope/core/management/populate/datasets.json
@@ -6,15 +6,15 @@
         "content": "HTA9_1_BA_F_ROI02.ome.tif"
     },
     {
-        "name": "HTA9_1_BA_F_ROI02 (Copy)",
+        "name": "HTA9_1_BA_F_ROI03",
         "id": 2001,
-        "description": "This is the test dataset from the VANDY Explorer",
-        "content": "HTA9_1_BA_F_ROI02.ome.tif"
+        "description": "VANDY test dataset 2 (https://styx.neurology.emory.edu/girder/#item/60e84a1868f4fe34fc7eef75)",
+        "content": "HTA9_1_BA_F_ROI03.ome.tif"
     },
     {
         "name": "HTA9_1_BA_F_ROI04",
         "id": 2002,
-        "description": "VANDY test dataset 2 (https://styx.neurology.emory.edu/girder/#item/60e84a1968f4fe34fc7eef79)",
+        "description": "VANDY test dataset 3 (https://styx.neurology.emory.edu/girder/#item/60e84a1968f4fe34fc7eef79)",
         "content": "HTA9_1_BA_F_ROI04.ome.tif"
     },
     {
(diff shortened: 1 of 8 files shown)


atlascope/core/management/populate/datasets.json
```diff
@@ -1,28 +1,34 @@
...
+        "description": "VANDY test dataset 1 (https://styx.neurology.emory.edu/girder/#item/60e84a1868f4fe34fc7eef71)",
         "content": "HTA9_1_BA_F_ROI02.ome.tif"
     },
     {
-        "name": "Local Vandy Dataset (Copy)",
-        "id": 202,
+        "name": "HTA9_1_BA_F_ROI02 (Copy)",
+        "id": 2001,
         "description": "This is the test dataset from the VANDY Explorer",
         "content": "HTA9_1_BA_F_ROI02.ome.tif"
```
annehaley: I don't think you need to make a copy dataset for this (where they are both local and point to the same file). Instead, just reference the first one in both investigations

atlascope/core/management/populate/datasets.json
```diff
@@ -1,28 +1,34 @@
 [
     {
-        "name": "Local Vandy Dataset",
-        "id": 201,
-        "description": "This is the test dataset from the VANDY Explorer",
+        "name": "HTA9_1_BA_F_ROI02",
```
annehaley, line 3: Can we apply a name to these that describes the content of the image somehow? And that makes the difference between the two tif datasets easier to discern?

atlascope/core/management/populate/datasets.json
```diff
@@ -1,28 +1,34 @@
...
-        "description": "This is the test dataset from the VANDY Explorer",
+        "name": "HTA9_1_BA_F_ROI04",
+        "id": 2002,
+        "description": "VANDY test dataset 2 (https://styx.neurology.emory.edu/girder/#item/60e84a1968f4fe34fc7eef79)",
+        "content": "HTA9_1_BA_F_ROI04.ome.tif"
+    },
+    {
+        "name": "HTA9_1_BA_F_ROI04 (remote)",
+        "id": 2003,
+        "description": "Vandy test dataset 1 (https://styx.neurology.emory.edu/girder/#item/60e84a1868f4fe34fc7eef71), retrieved via Vandy importer",
```
annehaley: Change the description on this one, it is a remote copy of test dataset 2.

atlascope/core/management/populate/embeddings.json
```diff
@@ -1,16 +1,16 @@
...
     {
-        "investigation": "Whole Team Investigation",
-        "id": 302,
-        "parent": "Local Vandy Dataset",
-        "child": "Local Vandy Dataset (Copy)",
-        "child_bounding_box": [100, 100, 200, 200]
+        "investigation": "Semantic zoom playground",
+        "id": 3001,
+        "parent": "HTA9_1_BA_F_ROI02",
+        "child": "HTA9_1_BA_F_ROI02 (Copy)",
```
annehaley: I think for the purpose of the Semantic zoom playground investigation, we should avoid putting the same image inside itself. Perhaps put dataset B inside dataset A in two different ways

atlascope/core/management/populate/datasets.json
annehaley: Is it possible to get a third unique dataset in here? That would make the embeddings playground that much more impressive

annehaley (changes requested):

atlascope/core/management/populate/datasets.json
waxlamp: I think I made the copy so I could embed two copies of the dataset in one "host"; if you're suggesting we get a third dataset for that purpose, I think that's a good idea. Will do.

atlascope/core/management/populate/datasets.json
waxlamp: I don't think it's a good idea, just because of the confusion in "provenance" it will cause when we go back to the source to explore other data.

atlascope/core/management/populate/datasets.json
waxlamp: See 9561caefcab8b5d5eb3c4875d8b4b820ae9b8d3.

atlascope/core/management/populate/datasets.json
waxlamp: See 336a697b710c836bf74cb689b265028c85d47a61.

atlascope/core/management/populate/embeddings.json
waxlamp: I fixed this by plopping the third dataset in there instead. See 9561caefcab8b5d5eb3c4875d8b4b820ae9b8d3d.

annehaley (approved):

naglepuff (approved):

PR #119, opened 2022-03-30
Convert images upon ingest

annehaley:
As a rival solution to #118, this branch explores the viability of converting large images upon ingest to a format supported by the `GDALFileTileSource` class (so that we can utilize `vsicurl` and avoid downloading the whole file)

I was able to use large-image-converter successfully, but when I do so, we lose some information from the metadata endpoint. Without conversion and using `OMETiffFileTileSource`, we get metadata like this, which includes `channelMap`:
```
{
  "levels": 4,
  "size_x": 5001,
  "size_y": 5001,
  "tile_size": 1024,
  "additional_metadata": {
    "levels": 4,
    "sizeX": 5001,
    "sizeY": 5001,
    "tileWidth": 1024,
    "tileHeight": 1024,
    "magnification": 19.912385503783355,
    "mm_x": 0.0005022,
    "mm_y": 0.0005022,
    "frames": [
      {
        "LightPath": {},
        "ID": "Channel:0:0",
        "Name": "NUCLEI",
        "SamplesPerPixel": "1",
        "IndexC": 0,
        "Frame": 0,
        "Index": 0,
        "Channel": "NUCLEI"
      },
      ...
    ],
    "IndexRange": {
      "IndexC": 12
    },
    "IndexStride": {
      "IndexC": 1
    },
    "channels": [
      "NUCLEI",
      "CD4",
      "CD3",
      "PD1",
      "KI67",
      "CD8",
      "EOMES",
      "GRZB",
      "IDO",
      "TBET",
      "CD68",
      "CD45"
    ],
    "channelmap": {
      "NUCLEI": 0,
      "CD4": 1,
      "CD3": 2,
      "PD1": 3,
      "KI67": 4,
      "CD8": 5,
      "EOMES": 6,
      "GRZB": 7,
      "IDO": 8,
      "TBET": 9,
      "CD68": 10,
      "CD45": 11
    }
  }
}
```

When I do convert and use the same tile source code, we get metadata like this:
```
{
  "levels": 6,
  "size_x": 5001,
  "size_y": 5001,
  "tile_size": 256,
  "additional_metadata": {
    "levels": 6,
    "sizeX": 5001,
    "sizeY": 5001,
    "tileWidth": 256,
    "tileHeight": 256,
    "magnification": null,
    "mm_x": null,
    "mm_y": null,
    "frames": [
      {
        "Frame": 0,
        "Index": 0
      },
      {
        "Frame": 1,
        "Index": 0
      },
      ...
      {
        "Frame": 11,
[...]

PR #122, opened 2022-04-04
Local storage

### LINKED ISSUE #112: Fuse Mount for more efficient large file caching
@mcovalt Could you add details here about how this was done on RGD?

annehaley:
As a rival solution to #118, this branch explores the viability of switching to complete local storage (no use of S3FileField or minio). This completely eliminates the problem that #118 and #119 seek to solve. The problem for the other two solutions was that we are trying to efficiently access large remote files. With this PR, we can switch to completely local file access. 

NOTE: after merging this PR, everyone will need to do a database reset and docker-compose rebuild

Resolves #112


mcovalt:
@waxlamp You asked earlier in the meeting if there are any cons to not using S3. I should have mentioned that Heroku's local file system is ephemeral from what I can tell. I bet this was the original reason Girder 4 uses MinIO. There are alternative ways to host this application so that we have persistent storage, but this would remove Heroku as a contender. The application is not on Heroku at the moment so I don't really see this as a problem.

atlascope/core/rest/endpoints/tile_endpoints.py
```diff
@@ -26,14 +27,11 @@ class TileMetadataView(GenericAPIView, mixins.RetrieveModelMixin):
 
     def get(self, *args, **kwargs):
         dataset = self.get_object()
-        cached = fsspec.open_local(
-            f'simplecache::{dataset.content.url}',
-            filecache={'cache_storage': '/tmp/files'},
-        )
+        content_location = Path(settings.BASE_DIR, dataset.content.name)
```
mcovalt:
```suggestion
        content_location = dataset.content.path
```

[I think this is given to us by Django under this property.](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.fields.files.FieldFile.path)

atlascope/settings.py
```diff
@@ -18,6 +18,8 @@ class AtlascopeMixin(ConfigMixin):
     ROOT_URLCONF = 'atlascope.urls'
 
     BASE_DIR = Path(__file__).resolve(strict=True).parent.parent
+    DATASET_STORAGE_DIR = Path('data')
```
mcovalt:
```suggestion
    MEDIA_ROOT = BASE_DIR / Path('data')
```

Two things are changed here:
1. Uses [`MEDIA_ROOT`](https://docs.djangoproject.com/en/4.0/ref/settings/#std:setting-MEDIA_ROOT).
2. Ensuring it's referenced as an absolute path.

atlascope/core/models/dataset.py
```diff
@@ -15,7 +14,7 @@ class Dataset(TimeStampedModel, models.Model):
     id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
     name = models.CharField(max_length=255)
     description = models.TextField(max_length=5000, blank=True)
-    content = S3FileField(null=True)
+    content = models.FileField(null=True, upload_to=settings.DATASET_STORAGE_DIR)
```
mcovalt:
```suggestion
    content = models.FileField(null=True)
```

This will default to [`MEDIA_ROOT`](https://docs.djangoproject.com/en/4.0/ref/settings/#std:setting-MEDIA_ROOT).

dev/django.Dockerfile
```diff
@@ -40,8 +40,6 @@ RUN pip install \
 # Use a directory name which will never be an import name, as isort considers this as first-party.
 WORKDIR /opt/django-project
 
-# Setup nginx to proxy localhost:9000 to minio:9000
-COPY ./dev/nginx.conf /etc/nginx/nginx.conf
 COPY ./dev/docker-entrypoint.sh /docker-entrypoint.sh
 RUN chmod +x /docker-entrypoint.sh
 ENTRYPOINT ["/docker-entrypoint.sh"]
```
mcovalt: We should also remove this entry point and the bash file it's using.

mcovalt: Additionally, we can remove the `nginx` installation on line 21.

mcovalt (changes requested):
Looks great! I added a few notes. Some were related to using more standard Django patterns to store the file. A couple were related to removing some dead code now that Nginx is gone.

dev/django.Dockerfile
annehaley: Changed in [055bbe6](https://github.com/atlascope/atlascope/pull/122/commits/055bbe606f991f7f5ef5eb570ab568d4f3bee47f)

mcovalt (approved):

PR #123, opened 2022-04-08
Integrate embeddings view into main investigation view

mcovalt:
This PR removes the `EmbeddingsDetail` view and integrates its functionality into the main `InvestigationsDetail` view. A few miscellany changes accompanied this move:
- The graph embedding traversal now includes a [check to prevent cycles](https://github.com/atlascope/atlascope/blob/3c9fd383cc320087074a502d08861ef55b89be44/web/src/views/InvestigationDetail.vue#L268). This, combined with the manual selection of the root dataset, should eliminate crashes caused by attempting to view an impossible or cyclic graph embedding.
- The `activeDataset` has been renamed to `rootDataset` throughout the codebase. This more clearly represents the state stored in the variable.
- The target for typescript was changed from ES5 to ES6 so I could use the spread operator.

To test:
- Ensure that changing the dropdown of dataset selector changes the depth of the initial view of the tree such that the selected layer becomes the new root.
- Ensure that the capability to view frames and pins are still available on the root dataset as they were on the single layer before.


naglepuff (approved):
These changes look great! I couldn't find anything that was worth holding this up. I did open #125 and #126 to track the additional changes that we discussed earlier today.

PR #124, opened 2022-04-08
Toggle note pins on click, not hover

naglepuff:
This PR updates some pin behavior on the front end. 

Previously, no pins were selected by default, and a user could select pins from the Pin List in the sidebar. Additionally, a user could hover over one pin on the map to view its note contents.

Now, when an investigation is loaded, all pins that can be selected are selected by default. This means that pins are rendered as features on the geojs map automatically. Users can still unselect pins. A selectable pin is defined (for now) as a pin whose parent dataset is the currently active dataset. Pin notes are now toggled by clicking on the pin in the map. This allows viewing several different notes at once.

This PR is strongly in the camp of "function over form," and as a result pin notes are not super pretty. I think a follow up PR could address this in the future.

In order to test this with the test data created by the populate script, you will need to open an investigation with pins that contain notes, and whose parent dataset is part of the investigation. That dataset will need to be selected as the active/root dataset. Upon loading such a dataset:

1. Verify that you see the pins as point annotations on the map
2. Verify that all pins that are selectable are shown
3. Click the pins
4. Verify that a card shows up with the clicked pin's note text
5. Click those pins again
6. Verify that the card with the note text is hidden
7. Zoom/pan on the map
8. Verify that the note cards move accordingly
9. Zoom/pan such that some pins move outside of the view window
10. Verify that cards belonging to pins who are no longer visible are hidden
11. Open the sidebar and deselect some pins
12. Verify that the pins and the note cards are hidden.

### DIFF
--- a/web/src/views/InvestigationDetail.vue
+++ b/web/src/views/InvestigationDetail.vue
@@ -238,10 +238,10 @@ export default defineComponent({
 
     function movePinNoteCards() {
       if (!featureLayer || !pinFeature || !map.value) { return; }
-      pinFeature.data().forEach((pinPoint: any) => {
-        const { x, y } = pinPoint;
+      pinFeature.data().forEach((pin: Pin) => {
+        const { x, y } = postGisToPoint(pin.child_location) || { x: 0, y: 0 };
         const newScreenCoords = pinFeature.featureGcsToDisplay({ x, y });
-        const note = pinNotes.value.find((pinNote) => pinNote.id === pinPoint.id);
+        const note = pinNotes.value.find((pinNote) => pinNote.id === pin.id);
         const {
           left, top, width, height,
         } = map.value?.getBoundingClientRect() || {
@@ -467,22 +467,6 @@ export default defineComponent({
       }));
     }
 
-    function getPinsToDisplay() {
-      // TODO: as we move towards embedding multiple datasets into the view,
-      // we will need a more sophisticated way to determine which pins to render
-      // and determining where they should be rendered
-      const pinFeatureData = selectedPins.value.map((pin) => {
-        const location: Point = postGisToPoint(pin.child_location) || { x: 0, y: 0 };
-        return {
-          ...location,
-          id: pin.id,
-          color: pin.color,
-          note: pin.note,
-        };
-      });
-      return pinFeatureData;
-    }
-
     watch(selectedPins, (newPins, oldPins) => {
       if (!featureLayer) {
         featureLayer = createLayer('feature', { features: ['point', 'line', 'polygon'] });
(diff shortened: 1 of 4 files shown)


web/src/views/InvestigationDetail.vue
```diff
@@ -222,6 +236,32 @@ export default defineComponent({
...
+            || note.notePositionX < left
+            || note.notePositionY < top) {
+          note.inBounds = false;
+        } else {
+          note.inBounds = true;
+        }
+      });
+    }
+
+    const debounceMovePinNoteCards = debounce(movePinNoteCards, 50);
```
mcovalt: I'm curious about why this is needed. I just read up on the concept. [Here's a great blog post](https://stoilsky.com/js-debounce-function/). It's probably common knowledge in the JavaScript world, but I learned something new. Is the callback function too heavy to do without this debounce?

mcovalt (approved):
Works great! This is getting very cool.

We're starting to leak quite a bit of `any` out of `useGeoJS`, but that's OK. Maybe at some point when we have time we can go through and change things up to prevent that. That effort is simply so we can have a good file to copy/paste into other GeoJS based projects. Not necessary.

naglepuff:
> We're starting to leak quite a bit of any out of useGeoJS, but that's OK. Maybe at some point when we have time we can go through and change things up to prevent that. That effort is simply so we can have a good file to copy/paste into other GeoJS based projects. Not necessary.

Would the fix for this `any` leakage be creating interfaces in `useGeoJS.ts` for things like layers and features? I can open an issue to track this, and I  want to understand what the best way forward is.

web/src/views/InvestigationDetail.vue
naglepuff: It's probably not necessary. I get nervous when doing DOM manipulation, as my understanding is that its slow, although maybe I'm wrong, or being overly cautious here.

Not debouncing doesn't seem to have an impact on performance at the moment, and the UX gain is very noticeable, so I'll take that out.

mcovalt:
I'm not really sure what The Right Way™ is. And I'm not sure it's very straightforward. We might need to get a bit creative.

One small example. Two ways to do the same task.
1. exporting `zoomLevel` as a [ref](https://vuejs.org/api/reactivity-core.html#ref) and then allowing components to [watch](https://vuejs.org/api/reactivity-core.html#watch) that ref.
2. exporting and typing the GeoJS map object and using its callback infrastructure to trigger events when the zoom changes.

I think the first is a more natural Vue pattern. That's all fine and good for the GeoJS map object. What about the layer(s)? We've discussed a bit on how to handle this, but I don't think we've come up with a good plan yet.

web/src/views/InvestigationDetail.vue
naglepuff: See 4d9a36a

naglepuff:
Opened #128 to keep track of the issue brought up regarding poor usage of `useGeoJS`.

mcovalt:
Ah, yeah. Removing the debounce makes it much smoother. It at least _feels_ faster [0]. Still glad I learned of the concept.

Of course, I'm viewing it on a very powerful machine, so I have no idea what this feels like on a "normal" computer.

[0] I remember reading about how Chrome does things that are bad for performance (in terms of watts and/or memory) in order for better _perceived_ performance. It's interesting to consider this concept of perceived performance.  I'm looking for the original source, [here's something similar](https://blog.mozilla.org/performance/2021/08/05/performance-in-progress/).

PR #127, opened 2022-04-11
Add widget for creating subimage datasets

annehaley:


web/src/store/index.ts
```diff
@@ -92,24 +102,24 @@ const {
...
         commit.setInvestigations(investigations.results);
       } else {
         commit.setInvestigations([]);
       }
     },
     async fetchCurrentInvestigation(context, investigationId: string) {
-      const { commit } = rootActionContext(context);
-      if (store.state.axiosInstance) {
-        const investigation = (await store.state.axiosInstance.get(`/investigations/${investigationId}`)).data;
+      const { commit, state } = rootActionContext(context);
```
naglepuff, line 116: Neat! I didn't know about this.

web/src/components/DatasetSubimageSelector.vue
```diff
@@ -0,0 +1,84 @@
...
+  setup() {
+    const selection = computed(() => store.state.subimageSelection);
+    const selectionMode = computed(() => store.state.selectionMode);
+    const savedNotification = ref<string>();
+
+    async function saveSubimageDataset() {
+      try {
+        const response = await store.dispatch.createSubimageDataset(selection.value);
+        if (response) savedNotification.value = 'Saved!';
+        else savedNotification.value = 'Failure to Save.';
```
naglepuff:
```suggestion
        savedNotification.value = response ? 'Saved!' : 'Failed to Save';
```

web/src/views/InvestigationDetail.vue
```diff
@@ -178,12 +184,14 @@ export default defineComponent({
     const selectedDataset: Ref<Dataset | null> = ref(null);
     const rootDataset = computed(() => store.state.rootDataset);
     const tilesourceDatasets = computed(() => store.getters.tilesourceDatasets);
+    const selectionMode = computed(() => store.state.selectionMode);
     /* eslint-disable */
+    let annotationLayer: any;
```
naglepuff:
```suggestion
    let selectionLayer: any;
```
Since this variable is tied to the functionality for selecting a subimage, I think this name better communicates its purpose.

web/src/components/DatasetSubimageSelector.vue
```diff
@@ -0,0 +1,84 @@
...
+      } catch (e) {
+        savedNotification.value = 'Failure to Save.';
+      }
+      store.commit.setSelectionMode(false);
+      setTimeout(() => { savedNotification.value = ''; }, 3000);
+    }
+    function selectionText() {
+      if (!selection.value) return '';
+      return `(${selection.value[0]}, ${selection.value[1]}) -> (${selection.value[2]}, ${selection.value[3]})`;
```
naglepuff:
```suggestion
      return `(${selection.value[0]}, ${selection.value[1]}), (${selection.value[2]}, ${selection.value[3]})`;
```

Super nitpicky I know but I think the comma here is slightly clearer.

web/src/components/DatasetSubimageSelector.vue
```diff
@@ -0,0 +1,84 @@
...
+      >
+        Cancel
+      </v-btn>
+      <v-btn
+        v-else
+        color="primary"
+        @click="saveSubimageDataset"
+      >
+        Create Subimage Dataset {{ selectionText() }}
+      </v-btn>
```
naglepuff:
```suggestion
      <v-btn
        color="secondary"
        @click="() => store.commit.setSelectionMode(false)"
      >
        Cancel
      </v-btn>
      <v-btn
        v-if="selection"
        color="primary"
        @click="saveSubimageDataset"
      >
        Create Subimage Dataset {{ selectionText() }}
      </v-btn>
```
I think the cancel button should still be enabled if there's a selection. That way a user can cancel creating the subimage if they've picked the "wrong" region.

web/src/components/DatasetSubimageSelector.vue
```diff
@@ -0,0 +1,84 @@
...
+        <v-btn
+          icon
+          v-bind="attrs"
+          v-on="on"
+          @click="() => store.commit.setSelectionMode(true)"
+        >
+          <v-icon>mdi-select</v-icon>
+        </v-btn>
+      </template>
+      <span>Select a portion of the data</span>
```
naglepuff:
```suggestion
      <span>Create a dataset for a region of the image</span>
```
I think the tooltip should at least mention creating a new dataset.

web/src/store/index.ts
```diff
@@ -27,6 +27,8 @@ export interface State {
     datasetEmbeddings: DatasetEmbedding[];
     datasetTileMetadata: { [key: string]: TileMetadata };
     rootDatasetFrames: TiffFrame[];
+    selectionMode: boolean;
+    subimageSelection: number[] | null;
```
naglepuff, line 31: I think these should be reset to false/null when the root dataset changes. If I select a region and then change the root dataset from the drop down, I still see the button prompting me to save, and I think this should effectively cancel the selection.

I see similar behavior with the following workflow: Investigate an investigation -> Click the subimage selection button -> select a region -> go back to investigation list view -> investigate an investigation. The button to save the subimage is still there, but the selection is not, and I may be in an entirely different investigation.

naglepuff (changes requested):

web/src/views/InvestigationDetail.vue
annehaley: Changed in [ef1705a](https://github.com/atlascope/atlascope/pull/127/commits/ef1705a3ef6f7106f6f27a43c6c39e25bbd792f7)

web/src/store/index.ts
annehaley: Changed in [3624b38](https://github.com/atlascope/atlascope/pull/127/commits/3624b38bb27b2e690b1bc7f082c81266042f1462)

naglepuff (approved):
Looks good to me!

PR #129, opened 2022-04-14
Add non-tiled image pins

### LINKED ISSUE #132: Support non-tiled image overlay pins

naglepuff:
Closes #132 

This PR adds a new type of dataset that can be used alongside pins: `non_tiled_image`.

Datasets of this type should have a file as their content, and that file should contain an image that is not large enough to justify tiling. 

Back-end changes include this new value for the `dataset_type` field, and some more data in the populate script.

Front end changes include support for displaying pins of this new type, using geojs quad features. In order to see and test these changes, switch to the `image-overlay-pins` branch and re-run the populate script to get some pins with this kind of data.

Open the Pin playground investigation and interact with the various pins. There are note-only pins and non-tiled image pins. Verify that they behave as you would expect (click a pin to show/hide the image, select/deselecting pins is not broken).


atlascope/core/rest/endpoints/dataset_endpoints.py
```diff
@@ -49,3 +61,15 @@ def subimage(self, request, pk):
...
+    @swagger_auto_schema(
+        responses={200: 'Image file', 404: 'Content not found'},
+    )
+    @action(detail=True, methods=['GET'], renderer_classes=[ContentRenderer])
+    def content(self, request, pk):
+        dataset = self.get_object()
+        image = PIL.Image.open(dataset.content.path)
+        buf = io.BytesIO()
+        image.save(buf, format='PNG')
+        return Response(buf.getvalue())
```
banesullivan: What if the image is a large tif?

You could universally use the `thumnail.png` endpoint from django-large-image in #135 

Or if these datasets should have their own endpoints, I'd maybe just use large_image here to produce the thumbnail rather than PIL so that it can handle any Dataset image content

naglepuff: Would this work for images that aren't tifs? It seems that if I hit the thumbnail endpoint for a Dataset whose contents are just a small .png image, I get a 404.

The goal of this PR is to be able to display non-tiled images  in geojs on the client (support for tiled images will come later). To that end I've created a new datasaet type, `non_tiled_image`. I could modify this endpoint so that it only works for datasets of that type.

When we were still using MinIO in the stack, I didn't need an endpoint at all, since the `content` field of such dataset objects was already the URL that I needed to get the image, but now that we've switched to local storage, I can't use the `content` field straight up.

banesullivan: > Would this work for images that aren't tifs?

We would need to add the `large_image_source_pil` dependency (which is lightweight).

> The goal of this PR is to be able to display non-tiled images in geojs on the client (support for tiled images will come later). 

django-large-image is built to handle this! Is the `content` in a `Dataset` always image data?

banesullivan: IMO, you should handle *all* image data through `django-large-image` and through tile serving. If you have a simple PNG, then there is only 1 Z layer and one tile.

Here is an example of django-large-image with a simple picture of Afie (the goodest boy there ever was): https://django-large-image.herokuapp.com/image-file/2/viewer

Tiles are served through django-large-image and it is viewed in GeoJS like any image
Screen Shot 2022-04-15 at 12 30 35 PM

naglepuff: Take a look at 60cca3f and let me know if I'm on the right track. I'm not using `django-large-image` yet (but if that PR moves before this one, I'll make any necessary changes). I agree that doing all image serving from the same place is the best thing to do. Thanks for pointing me in the right direction.

(very handsome dog btw)

banesullivan: Those changes look great!

web/src/views/InvestigationDetail.vue
```diff
@@ -464,17 +473,78 @@ export default defineComponent({
...
+      if (existingOverlay) {
+        nonTiledOverlayFeature.data(
+          // eslint-disable-next-line
+          nonTiledOverlayFeature.data().filter((overlay: any) => overlay.pinId !== pin.id),
+        ).draw();
+        nonTiledOverlayFeature.draw();
+      } else {
+        try {
+          const urlRoot = process.env.VUE_APP_API_ROOT;
+          const url = `${urlRoot}/datasets/${dataset.id}/tiles/0/0/0.png`;
```
banesullivan: Hm. I'd like to think about this one a bit and perhaps consult @manthey.

I'm wondering if I should add an endpoint in `django-large-image` that will pass the whole image.

Otherwise, I would recommend using a thumbnail here rather than the 0-level tile:

```suggestion
          const url = `${urlRoot}/datasets/tile_source/${dataset.id}/thumbnail.png`;
```

(but that change would only be valid AFTER #135 )

manthey: In the Girder 3 instance, thumbnail can specify a maximum size.  I don't recall if django-large-image supports this.

banesullivan: Just added that in django-large-image v0.4.4
Screen Shot 2022-04-20 at 11 52 47 AM

banesullivan: (please ignore the typo in `max_height`'s docs

waxlamp (commented):
This doesn't seem to be working for me. When I click on an image overlay pin, I see that the image itself is retrieved (in the network tab of the dev console) but nothing appears in-app.

naglepuff:
@waxlamp I'll take a look and see what's going on, however I'd like for #135 to merge in first, and use django-large-image to fully resolve a conversation above. That PR seems about ready to go.

naglepuff:
@waxlamp Give it another shot with 7e068b5. If you're still running into trouble I'd be happy to work together to figure out whatever is going one.

web/src/views/InvestigationDetail.vue
naglepuff: I've applied this change in 7e068b5

atlascope/core/rest/endpoints/tile_endpoints.py
```diff
@@ -12,14 +13,17 @@
...
+        content__isnull=False, dataset_type__in=['tile_source', 'non_tiled_image']
+    )
     serializer_class = DatasetSerializer
 
     default_colors = ['#ffffff']
 
     def get_path(self, *args, **kwargs) -> str:
         """Retreive path to image file from data record."""
         dataset = self.get_object()
+        print(dataset)
```
waxlamp:
```suggestion
```

web/src/utilities/utiltyFunctions.ts
```diff
@@ -13,7 +13,17 @@ export function postGisToPoint(location: string): Point | undefined {
       y: parseInt(yPart.substring(0, yPart.length - 1), 10),
     };
   } catch (error) {
+    // eslint-disable-next-line
```
waxlamp: What error are you getting from this?

Perhaps a better way in general to signal errors is to `throw new Error('message')`.

web/src/views/InvestigationDetail.vue
```diff
@@ -488,8 +489,14 @@ export default defineComponent({
     });
 
     const selectedPins: Ref<Pin[]> = computed(() => store.state.selectedPins);
+    /**
+     * Create note cards for any pins without a child dataset.
+     */
```
waxlamp: Use C++ style comments.

```suggestion
    // Create note cards for any pins without a child dataset.
```

web/src/views/InvestigationDetail.vue
```diff
@@ -488,8 +489,14 @@ export default defineComponent({
...
     function createPinNotes() {
-      pinNotes.value = store.state.currentPins.map((pin) => ({
+      const noteOnlyPins = store.state.currentPins.filter(
+        (pin: Pin) => (!pin.child && pin.note?.length),
+      );
```
waxlamp, line 505: I thought we had discussed giving pins an explicit type field? That would make this type of filtering easier to understand in code, and easier to carry out at runtime.

web/src/views/InvestigationDetail.vue
```diff
@@ -498,17 +505,84 @@ export default defineComponent({
       }));
     }
 
+    async function toggleNonTiledImageOverlay(pin: Pin, dataset: Dataset) {
+      if (!dataset.content) return;
+      if (!featureLayer) return; // sanity check. this should exist
```
waxlamp: In both of these cases, if the check represents an impossible condition (placed into an if statement to satisfy the typechecker) then it is probably better to explicitly throw an `Error` object rather than silently `return`ing.

It's also a best practice, stylewise, to place single statement blocks inside curly brackets, to aid in legibility and prevent accidental problems when adding more statements to the body of such a block.

web/src/views/InvestigationDetail.vue
```diff
@@ -498,17 +505,84 @@ export default defineComponent({
...
+      if (!nonTiledOverlayFeature) {
+        nonTiledOverlayFeature = featureLayer.createFeature('quad');
+      }
+      const quadData = nonTiledOverlayFeature.data() || [];
+      // eslint-disable-next-line
```
waxlamp: What is the eslint error here?

web/src/views/InvestigationDetail.vue
```diff
@@ -498,17 +505,84 @@ export default defineComponent({
...
+      const existingOverlay = quadData.find((overlay: any) => overlay.pinId === pin.id);
+      if (existingOverlay) {
+        nonTiledOverlayFeature.data(
+          // eslint-disable-next-line
```
waxlamp: How about here? (I assume it's the same error as above?)

web/src/views/InvestigationDetail.vue
```diff
@@ -498,17 +505,84 @@ export default defineComponent({
...
+          nonTiledOverlayFeature.data().filter((overlay: any) => overlay.pinId !== pin.id),
+        ).draw();
+        nonTiledOverlayFeature.draw();
+      } else {
+        try {
+          const urlRoot = process.env.VUE_APP_API_ROOT;
+          const url = `${urlRoot}/datasets/tile_source/${dataset.id}/thumbnail.png`;
+          const image: HTMLImageElement = await getNonTiledImage(url);
+          image.crossOrigin = 'use-credentials';
+          if (store.state.axiosInstance) {
```
waxlamp: I would invert this test and place it at the top of the function, so you can early exit if `axiosInstance` doesn't exist yet, and you don't have to complexify the logic here with a type narrowing.

web/src/views/InvestigationDetail.vue
```diff
@@ -498,17 +505,84 @@ export default defineComponent({
...
+          // eslint-disable-next-line
+          nonTiledOverlayFeature.data().filter((overlay: any) => overlay.pinId !== pin.id),
+        ).draw();
+        nonTiledOverlayFeature.draw();
+      } else {
+        try {
+          const urlRoot = process.env.VUE_APP_API_ROOT;
+          const url = `${urlRoot}/datasets/tile_source/${dataset.id}/thumbnail.png`;
+          const image: HTMLImageElement = await getNonTiledImage(url);
+          image.crossOrigin = 'use-credentials';
```
waxlamp: I am curious whether the problems we were seeing with the overlays not appearing in Chrome and Brave were due to the type narrowing if statement below, and this could remain `anonymous`. That is probably worth a check.

web/src/views/InvestigationDetail.vue
```diff
@@ -498,17 +505,84 @@ export default defineComponent({
...
+          if (store.state.axiosInstance) {
+            const imageMetadata = await (await store.state.axiosInstance.get(`${urlRoot}/datasets/tile_source/${dataset.id}/metadata`)).data;
```
waxlamp: Do you truly need two nested `await`s here?

web/src/views/InvestigationDetail.vue
```diff
@@ -498,17 +505,84 @@ export default defineComponent({
...
+            const ul = postGisToPoint(pin.child_location) || { x: 0, y: 0 };
+            const lr = {
+              x: ul.x + (imageMetadata.sizeX || 0),
+              y: ul.y + (imageMetadata.sizeY || 0),
```
waxlamp: Since you already have these values from the `Image` object that contains the actual image data, I wonder if we can avoid the extra call to the metadata endpoint here (this is if the Chrome/Brave issue was solved by the type narrowing, rather than the use of the `Image` object).

web/src/views/InvestigationDetail.vue
```diff
@@ -498,17 +505,84 @@ export default defineComponent({
...
+              ul,
+              lr,
+              image,
+              pinId: pin.id,
+            });
+            nonTiledOverlayFeature.data(quadData).draw();
+          }
+        } catch (error) {
+          // eslint-disable-next-line
+          console.error(`Failed to show overlay for dataset ${dataset.id}.`);
```
waxlamp: You can throw an `Error` object to avoid the eslint disabling line.

web/src/views/InvestigationDetail.vue
```diff
@@ -498,17 +505,84 @@ export default defineComponent({
...
+        }
+      }
+    }
+
+    function toggleDatasetPin(pin: Pin) {
```
waxlamp, line 561: What's the purpose of this function? Why does it have a single-case switch statement (rather than just an if statement)?

web/src/views/InvestigationDetail.vue
```diff
@@ -498,17 +505,84 @@ export default defineComponent({
...
+      switch (childDataset.dataset_type) {
+        case 'non_tiled_image':
+          toggleNonTiledImageOverlay(pin, childDataset);
+          break;
+        default:
+          break;
+      }
+    }
+
     watch(selectedPins, (newPins, oldPins) => {
```
waxlamp, line 581: Why `watch` the selection state, rather than attach logic to the toggle state per pin?

(This is really an architecture question--I'm not necessarily suggesting any changes here.)

waxlamp (changes requested):
The main change I'm requesting is in cleaning up the "fix" we put in during pair programming for the tiled images not showing up in Chrome/Brave. Please take a look at the cleanup suggestions there and see if things continue working in both Firefox and Chrome.

And getting @manthey's eyes on what might be happening here would be a good idea too (since it's happening internally in GeoJS, but showing an external difference between browser engines).

web/src/views/InvestigationDetail.vue
naglepuff: We have discussed this, and I'm not convinced we get all that much from such a field. I mostly think it would be redundant, since `pin.child.dataset_type` already exists. I could be talked into creating this field, but I don't see a need for it now.

web/src/views/InvestigationDetail.vue
naglepuff: It doesn't like the `any` in `filter((overlay: any), ...)`

web/src/views/InvestigationDetail.vue
naglepuff: Same as above.

web/src/views/InvestigationDetail.vue
naglepuff: I was thinking ahead to supporting other `dataset_type`s. For example, pins that display tiled data would have a different `dataset_type`, and would be displayed differently than these small image pins. This function would take a pin and call the corresponding toggling function. The switch does look silly now with just one case, so I could remove this if you'd like. Just let me know.

web/src/views/InvestigationDetail.vue
naglepuff: Selecting/deselecting pins is done in the `PinList` component. Instead of bubbling up events to the `InvestigationDetail`, we're using the store to keep track of which pins the user has selected/deselected, so the `InvestigationDetail` can just monitor this state via a computed property and a watcher. Let me know if I'm answering your question or not.

web/src/views/InvestigationDetail.vue
naglepuff: No I do not

0ed1295

web/src/views/InvestigationDetail.vue
naglepuff: Something I noticed when testing this was that the issue with chrome seems to be that the height and width of the image element are not available right away. I'm at a loss as to why though... I managed to figure out that `image.loading` in Firefox is `eager`, and in Chrome it is `auto`. Manually setting `image.loading = 'eager'` did not seem to make a difference.

Essentially in Chrome, the height and width are not available at the time of calculating `ul` and `lr`. There was no problem loading the image (we could see that via the network tab), but the quad feature had a width and height of 0.

In 2ec5d2c I've removed the utility function, in favor of using the `/metadata` endpoint to get the image dimensions.

web/src/views/InvestigationDetail.vue
naglepuff: Fixed in 2ec5d2c

web/src/views/InvestigationDetail.vue
naglepuff: As stated in another comment, the issues with chrome were due to timing. Now, both `anonymout` and `use-credentials` work. I'm not sure which one is *best* here. Is there any reason we wouldn't want to send credentials?

web/src/views/InvestigationDetail.vue
naglepuff: In that case would it be better to remove the try/catch block? I don't know if re-throwing with a custom error message will be meaningful at this point. I've removed the try/catch 1fe6e72; I'm interested to head your opinion on that change.

web/src/views/InvestigationDetail.vue
naglepuff: see 1fe6e72

web/src/utilities/utiltyFunctions.ts
naglepuff: I've made it more clear what I'm trying to guard against in e229ef2

web/src/views/InvestigationDetail.vue
waxlamp: To summarize recent discussions: I think perhaps we are both wrong. We discussed on 2022-05-25 using Django more explicitly to model different types of pins, and I asked Mike to experiment with this idea on a branch based on this one. So, for the moment, this conversation can be resolved.

web/src/views/InvestigationDetail.vue
waxlamp: It might seem silly, but crafting the precise type that your code expects to use here is generally a good idea. Over time, this custom type of `overlay` will grow to reflect what GeoJS expects of it, as your code sees it. It might feel tautological to do that, but there is some benefit that comes from that--that same benefit is being thrown away when you type `overlay` and `any` and ask the typechecker to look the other way.

It's sort of like fastening your seatbelt every time you drive a car, even if it's an edge case like, you're driving from one end of the parking lot to the other.

The analogy is not perfect because seatbelt fastening is quite low cost, and the potential benefits of that cost are quite high. But my larger point is, as you get better at using TypeScript, you can drive the cost of this kind of custom typing down (by developing good habits and practices) and you will accrue small amounts of value over time by doing so.

As an exercise, I'd like you to see how this kind of typing could be done here, with the aim of eliminating the eslint-disable statement (and thereby leaning into typechecking support).

web/src/views/InvestigationDetail.vue
waxlamp: See above comment.

web/src/views/InvestigationDetail.vue
waxlamp: I don't think there is.

web/src/views/InvestigationDetail.vue
naglepuff: See f4c19af. Is this what you had in mind?

web/src/views/InvestigationDetail.vue
naglepuff: f4c19af

web/src/views/InvestigationDetail.vue
waxlamp: The question is, what is the try/catch block doing there specifically?

According to the [`await` docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await) the value a promise is rejected with will be thrown. So, to be typesafe I think it's best to wrap the `await` expression in a try/catch in order to rethrow the value that arose, or other handle it. You could put the try/catch around a longer block of code but I think that choice depends on how you want to route control flow in case of a rejection promise.

Let me know if this makes sense--if so, I can help narrow down a design for where try/catch should be in this code.

web/src/views/InvestigationDetail.vue
waxlamp: I think maybe under the new design we've discussed, this might improve generally (for the record, the problem is not the single case; it's the need for a switch statement at all).

waxlamp (approved):
I think this is looking good, especially if we're attempting to revamp the design a bit in a followup branch.

Some of the commits in this PR mixed general improvements with specific fixes requested in my last review. As such, it's a bit hard for me to tell if all my comments have been addressed, so I'm asking you to make sure they have before merging (that is, I'm giving this a prospective approval, subject to your making sure you've addressed everything).

web/src/views/InvestigationDetail.vue
waxlamp: Yes, that looks great!

web/src/views/InvestigationDetail.vue
naglepuff: Resolved during meeting

web/src/views/InvestigationDetail.vue
naglepuff: [Here](https://axios-http.com/docs/handling_errors) is the relevant axios documentation. I've added a `.catch()` to the call to `toggleNonTiledImageOverlay`. If this function throws an error due to a response outside the 2XX range, it is because the associated dataset either has no content or its content couldn't be opened with `django-large-image`. An error is thrown with a specific message in this case. If an error is thrown due to no response or an inability to make the request, I simply re-throw that error. This change was made in dcc17c7

PR #135, opened 2022-04-14
Use django-large-image

banesullivan:
Look at all these new endpoints you now get for free!

Screen Shot 2022-04-14 at 4 22 17 PM

### Implementation

I removed the existing API Views you all implemented for tiling and added a new `GenericViewSet` using the `LargeImageDetailMixin`. To make this work with your existing design choices and front-end query parameters, I overrode 3 methods:

- `get_path`: this is necessary to override when using `django-large-image`. It grabs the file path from your model (if implementing FUSE, this is where you'd do it)
- `open_tile_source`: Upstream, this method typically uses `large_image.open` to open the files. However, since you all are dealing with OME tiffs and you have GDAL installed, the `GDALFileTileSource` class was used over `OMETiffFileTileSource`. Overriding this method enables you to choose your own logic for how the tile source class is determined.
- `get_style`: This builds a style dictionary for large-image following [large-image's specs](https://girder.github.io/large_image/tilesource_options.html#style). The original code you all implemented for frame compositing wasn't needed as the same functionality is handled directly by large-image.

I placed this viewset under `api/v1/datasets/tile_source` to separate it from the already existing viewset for `Dataset`.

_Why not combine this with your existing `DatasetViewSet` in `atlascope/core/rest/endpoints/dataset_endpoints.py`?_ Because the `queryset`s were different. I don't really know what's going on under the hood here, but I noticed that tile serving is only valid for certain `Dataset` records as captured by the `.filter(content__isnull=False, dataset_type='tile_source')` 

### Caveats

- The styling OpenAPI specs rendered in swagger are not valid for this project as we have to override the `get_style` method here.
- I did not include an admin widget viewer here because the UI for it does not currently handle multi-frames (coming soon)

Upstream changes to support this work: https://github.com/girder/django-large-image/compare/0.4.0...0.4.1


atlascope/core/rest/endpoints/tile_endpoints.py
```diff
@@ -1,135 +1,72 @@
...
+            colors = self.default_colors
+        style = {'bands': []}
+        for channel, color in list(zip(channels, cycle(colors))):
+            style['bands'].append(
+                {
+                    'frame': channel,
+                    'palette': ['#000', color],
+                }
+            )
+        return style
```
banesullivan: Ideally, we should not override this method, but instead, have the front-end create the JSON style per https://github.com/girder/django-large-image#-styling and https://girder.github.io/large_image/tilesource_options.html#style

banesullivan: This would also prevent opening the source twice - note that it isn't cached because the first open is without a style and the second is with a style

annehaley (approved):
I like the look and organization of this a lot more, it'll be nice to move some of that custom stuff out to a dependency.

atlascope/core/rest/endpoints/tile_endpoints.py
annehaley: @naglepuff Do you think you'd be able to accomplish this frontend refactoring? If so, it might be appropriate to make another PR that goes into this one

banesullivan: Merging this and having a follow up front-end PR seems appropriate to me

naglepuff: Yeah, for sure I can take that on

naglepuff: Just opened #139

setup.py
```diff
@@ -50,7 +50,7 @@
         'importlib_metadata>=3.6',
         'jsonschema',
         'large-image[gdal,ometiff]>=1.13.0',
-        'django-large-image>=0.4.1',
+        'django-large-image>=0.4.3',
```
banesullivan:
```suggestion
        'django-large-image>=0.4.4',
```

banesullivan:
@annehaley, good to merge?

annehaley:
> @annehaley, good to merge?

I believe so but I'll default to @waxlamp's opinion.

mcovalt (changes requested):
I love seeing all this extra functionality at a +60 / -179 diff! Unfortunately I'm getting errors on a fresh rebuild. I can look into it a bit more, but wanted to block merging until we got it figured out. It very well could be something I'm doing wrong!

mcovalt:
I added a test to show the errors I am getting.

banesullivan:
Fixed in django-large-image 0.5.0 - I had implemented the `open_tile_source` specifically for this project and made a tweak to it recently without updating here. 

----

On this topic: there is now a `source` parameter for `django-large-image` to control which tile source to use. So this could be replaced by doing:

```
/api/v1/datasets/tile_source/2000/tiles/metadata?source=ometiff
```

And not overriding `open_tile_source`

mcovalt (approved):
Works well now! Thanks @banesullivan for helping out with this.

annehaley (approved):

naglepuff (approved):

banesullivan:
I'll let one of you all hit the mighty green button - also not sure what your merging policy/pattern is (e.g. squash and merge?)

annehaley:
@waxlamp suggested that we should only merge once everyone has had a chance to test the changes on their own branch to ensure that it doesn't affect any existing work. So I'll just tag him and @marySalvi to bump this on their radar ;)

waxlamp (approved):
Tried this out and it's working for me. On top of that, the other group members are reporting no trouble (and some definite benefit) from this change.

I see that there are some conflicts with main at this point, but I don't know what they are. @banesullivan, do you have the bandwidth to resolve those conflicts? If not, @mcovalt, could you do it?

Thanks!

mcovalt:
I'll resolve the merge conflicts.

PR #136, opened 2022-04-15
Add yarn to docker-compose services

annehaley:
It may be beneficial to roll yarn into our docker services so that we don't need separate terminals running `docker-compose up` and `yarn serve`. @naglepuff @mcovalt What do you think?

### DIFF
--- a/dev/yarn.Dockerfile
+++ b/dev/yarn.Dockerfile
@@ -1,9 +1,0 @@
-FROM yarnpkg/node-yarn:latest
-
-RUN mkdir /web
-WORKDIR /web
-COPY ./web /web
-RUN npm cache clean -f
-RUN npm install n -g -q
-RUN n stable -q
-RUN npm install -q
--- a/dev/yarn.Dockerfile
+++ b/dev/yarn.Dockerfile
@@ -1,0 +1,9 @@
+FROM yarnpkg/node-yarn:latest
+
+RUN mkdir /web
+WORKDIR /web
+COPY ./web /web
+RUN npm cache clean -f
+RUN npm install n -g -q
+RUN n stable -q
+RUN npm install -q
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -38,6 +38,14 @@ services:
       - postgres
       - rabbitmq
 
+  yarn:
+    build:
+      context: .
+      dockerfile: ./dev/yarn.Dockerfile
+    volumes:
+      - ./web:/web
+    command: yarn serve
+
   demotiles:
     build:
       context: ./dev/demotiles
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -39,7 +39,7 @@ services:
       - rabbitmq
 
   yarn:
-    image: node:latest
+    image: node:18
     working_dir: /web
     volumes:
       - ./web:/web
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -39,14 +39,15 @@ services:
       - rabbitmq
 
   yarn:
-    build:
-      context: .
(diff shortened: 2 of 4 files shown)


banesullivan:
@annehaley, I think this should be done anytime a vue client is added to a Girder 4 project. Would you please write up a miniblog in Notion on how to do this for any Girder 4 project? - looks simple enough, but I'd love to capture it in documentation

annehaley:
Sure, I've never done that on Notion, though. Could you PM me and tell me how?

waxlamp:
This would be a good thing to discuss with the #girder4 channel too. It seems like a cool idea but I'm having trouble intuiting the pros and cons of the approach.

web/package.json
```diff
@@ -56,7 +56,9 @@
     "parserOptions": {
       "ecmaVersion": 2020
     },
-    "rules": {},
+    "rules": {
+      "no-console": "off"
```
marySalvi: I recommend still having a warning so console.log doesn't go unnoticed:

`    'no-console': ["warn", {
      allow: ["warn", "error"]
    }],`

dev/yarn.Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM yarnpkg/node-yarn:latest
+
+RUN mkdir /web
+WORKDIR /web
+COPY ./web /web
+RUN npm cache clean -f
+RUN npm install n -g -q
+RUN n stable -q
```
mcovalt: I think it makes more sense to use a base image with the correct Node.js version rather than use the [`n` tool](https://www.npmjs.com/package/n).

dev/yarn.Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+RUN npm install -q
```
mcovalt: This will install `node_modules` in the /web directory, but we seem to mount over that in docker-compose.

mcovalt (commented):
How does this handle changes to dependencies? Do we need to execute `yarn install` in the Docker container? Rebuild the image?

web/package.json
annehaley: Changed in [d3b8233](https://github.com/atlascope/atlascope/pull/136/commits/d3b82334e132ed913323c5afa983c4c00fd8d4f8)

dev/yarn.Dockerfile
annehaley: After my discussion with Daniel, it seems that for development, the better approach is to mount the local copy of the `node_modules` directory instead of using a dockerfile to build and install them separately inside the container. I changed this in [1dd90b0](https://github.com/atlascope/atlascope/pull/136/commits/1dd90b08f7ff7e536999263030583820c9eb0bd3).

dev/yarn.Dockerfile
annehaley: See above

docker-compose.override.yml
```diff
@@ -38,6 +38,17 @@ services:
       - postgres
       - rabbitmq
 
+  yarn:
+    image: node:latest
```
waxlamp: Can we pin to an image with the right major version?

waxlamp (approved):
I'm still uneasy with containerizing the frontend build process, but this seems to work well enough. I'm glad to keep this innovation in place and see how we like it. If it's useful to us we can share this with other projects in the group.

docker-compose.override.yml
annehaley: Yes; changed in [3889dec](https://github.com/atlascope/atlascope/pull/136/commits/3889dec9b0c9a6f633631bd441e20f90ca9b1dcb)

waxlamp: Thank you! My approval stands.

PR #140, opened 2022-04-19
Spawn subimage jobs

annehaley:


waxlamp (commented):
I've seen this working and it's a good improvement. When @mcovalt approves via a formal code review, then this will be ok with me as well.

web/src/components/JobPane.vue
```diff
@@ -0,0 +1,219 @@
...
+  </v-card>
+</template>
+
+<style scoped>
+.job-type-description {
+  color: gray;
+}
+</style>
+
+<script>
```
mcovalt: In other components we use TypeScript. @waxlamp Is this still something we want to continue requiring in Phase I?

atlascope/core/models/dataset.py
```diff
@@ -1,8 +1,15 @@
...
+from pathlib import Path
+import tempfile
+
 from django.conf import settings
 from django.contrib import admin
 from django.core.exceptions import ValidationError
 from django.db import models
+from django.dispatch import receiver
 from django_extensions.db.models import TimeStampedModel
+from large_image_converter import _output_tiff
```
mcovalt, line 11: I'm a little nervous about importing a "private" (prefixed by underscore) method. We should definitely pin the version of `large_image_converter` if we do this. Otherwise, is there an official public method we can use?

mcovalt:
I noticed a new [package-lock.json](https://github.com/atlascope/atlascope/blob/279a0caf2c04348c2489d2a0ba30f1e148e5e4d3/web/package-lock.json). Are we moving away from yarn in favor of npm? I think we should choose one or the other instead of holding onto multiple lock files.

annehaley:
> I noticed a new [package-lock.json](https://github.com/atlascope/atlascope/blob/279a0caf2c04348c2489d2a0ba30f1e148e5e4d3/web/package-lock.json). Are we moving away from yarn in favor of npm? I think we should choose one or the other instead of holding onto multiple lock files.

This was not intended, thank you for catching that. I must have used an npm install command somewhere along the way

atlascope/core/models/dataset.py
annehaley: this private function was all that I could find to do what we need. Some public functions invoke this but do much more additional stuff. The module as a whole has most of its functions as private with very narrow public use cases. I pinned the working version of `large_image_converter`, though. For now that will have to do. [24b2e61](https://github.com/atlascope/atlascope/pull/140/commits/24b2e610665820fb01468468b58239c2ebdaf749)

web/src/components/JobPane.vue
annehaley: The conversion is simple enough, we may as well continue typescript. Changed in [d238ad3](https://github.com/atlascope/atlascope/pull/140/commits/d238ad3d22371cf03a2fe54af101f51cbeee748e)

mcovalt (approved):
Awesome! Works well. This is a pretty massive PR and I did my best to understand the code changes.

I added a commit, 8f293fa969861917b691c9e71b9e82476c54be16, to change the default method pyvips used to shrink the image. It seems libvips on different architectures do not support the default 'nearest' method of shrinking an image. The method 'mode' should be more broadly supported and produce functionally equivalent results.

Take a look at that commit and maybe test things out once more before merging.

web/src/components/JobPane.vue
waxlamp: (Sticking with TypeScript is a good choice even for Phase I, because it requires a relatively small up-front investment of time to avoid relatively large time losses later.)

atlascope/core/models/dataset.py
waxlamp: @manthey, can you comment on this? Could it be possible to make `_output_tiff()` an official part of the API?

waxlamp:
> > I noticed a new [package-lock.json](https://github.com/atlascope/atlascope/blob/279a0caf2c04348c2489d2a0ba30f1e148e5e4d3/web/package-lock.json). Are we moving away from yarn in favor of npm? I think we should choose one or the other instead of holding onto multiple lock files.
> 
> This was not intended, thank you for catching that. I must have used an npm install command somewhere along the way

Friendly reminder not to do things like `git add *` and other similar command anti-patterns, like `git commit -a` without reviewing what was added (don't know if either of those happened in this case, but I figured I'd trot out another of my Git PSAs).

atlascope/core/models/dataset.py
manthey: It looks like you are really trying to do what `tifftools.commands.tiff_concat` does.  I haven't digested what you are doing besides concatenating a set of tiled tiff images.

annehaley: I opened https://github.com/atlascope/atlascope/pull/166 to use tiff_concat instead. It works well!

Issue #128, opened 2022-04-14
Better ultilize useGeoJS


naglepuff:
I'm not really sure what The Right Way™ is. And I'm not sure it's very straightforward. We might need to get a bit creative.

One small example. Two ways to do the same task.
1. exporting `zoomLevel` as a [ref](https://vuejs.org/api/reactivity-core.html#ref) and then allowing components to [watch](https://vuejs.org/api/reactivity-core.html#watch) that ref.
2. exporting and typing the GeoJS map object and using its callback infrastructure to trigger events when the zoom changes.

I think the first is a more natural Vue pattern. That's all fine and good for the GeoJS map object. What about the layer(s)? We've discussed a bit on how to handle this, but I don't think we've come up with a good plan yet.

_Originally posted by @mcovalt in https://github.com/atlascope/atlascope/issues/124#issuecomment-1099251993_

PR #141, opened 2022-04-20
Move pan and zoom GeoJS events to useGeoJS

naglepuff:
@mcovalt can you take a look at my attempt to move some geojs stuff back to `useGeoJS`? So far I've addressed zoom/pan but I'm open to expanding the scope of this PR if there's more stuff that we should address at this time.

I'll mention #128 here but this PR does not yet fully address the issue (there's still the matter of individual layers).


mcovalt (approved):
Yeah this is perfect. I say we should merge this in. We can tackle layers later when we have free time.

PR #143, opened 2022-04-25
Move geojs layers to `useGeoJS`

### LINKED ISSUE #128: Better ultilize useGeoJS

naglepuff:
Closes #128 

This PR creates composables `useGeoJSFeature` and `useGeoJSLayer`. These encapsulate GeoJS objects and their functionaliry, much like `useGeoJS` encapsulate map-level creation and operations. They are essentially thin wrappers around GeoJS functions with better typing. As we use more GeoJS functionality, these should be expanded to include those operations as well.


waxlamp:
@naglepuff, what's the status of this PR?

naglepuff:
@waxlamp I'm going to rebase to `main`, clean it up, and then pass it over to @mcovalt for review.

mcovalt (approved):
This is awesome. I've spent quite a bit of time playing around with this implementation and other ideas related to it. I think you've implemented the clearest approach.

PR #144, opened 2022-04-25
Use color picker for frame colors

naglepuff:
This PR makes the interface for changing frame colors more user friendly by adding a color swatch to the input field. This swatch is clickable, and brings up a color picker when clicked.

### DIFF
--- a/web/src/components/InvestigationDetailFrameMenu.vue
+++ b/web/src/components/InvestigationDetailFrameMenu.vue
@@ -1,88 +1,94 @@
 <template>
-  <v-menu
-    v-model="showFrames"
-    :close-on-content-click="false"
-    offset-y
-    max-height="500px"
-  >
-    <template
-      v-slot:activator="{ on: onMenu, attrs: attrsMenu }"
-    >
-      <v-tooltip bottom>
-        <template v-slot:activator="{ on: onTooltip, attrs: attrsTooltip }">
-          <v-btn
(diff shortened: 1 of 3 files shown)


annehaley (commented):
The code looks good, but in testing it out, I found that the channels didn't update with the new colors. Additionally, when I select a color on the edge of the color picker (dragging and releasing my mouse outside of the picker to do so, as shown in the video below), the whole dialog disappears. Is there a way you can modify the off-click of the dialog to only fire if the _start_ of the click is off the dialog?

**Edit:** I see the update button now, but is it possible to fire the update when a color is done being selected (i.e. when the singular picker is unfocused)?

https://user-images.githubusercontent.com/44912689/165541795-d9646d02-0346-4935-97e3-15acab28d4fe.mov

naglepuff:
The channels only update when the `Update` button is clicked. This change was made in #142.

To address your usability concern, check out f2300f5. I did some digging and couldn't find anything specific to the layout I had before, but I found some [similar issues](https://github.com/vuetifyjs/vuetify/issues/12461) with `v-dialog`s. Those issues were addressed in [this PR](https://github.com/vuetifyjs/vuetify/pull/13244), so I changed our UI a bit to get the desired functionality.

annehaley (approved):
That works much better and makes it more clear that the user needs to hit the update button. Thanks!

PR #145, opened 2022-04-26
Add zoom range for pins

### LINKED ISSUE #146: Incorporate "scale" into pins
As an enhancement to existing pin functionality, pins should be able to somehow specify at what scale they are useful. This information could be used by the client to show/hide pins at different scales (e.g. zoom levels)

naglepuff:
Closes #146 

This PR makes the following changes:

**Backend**
- Add `maximum_zoom` and `minimum_zoom` fields to the pin model
- Add more example pins to the populate script

**Frontend**
- Regenerate Atlascope typescript types
- Show/hide pins based on current zoom level


annehaley (approved):
We may have to merge migrations because now there is more than one pending PR with a migration. This all looks good though!

naglepuff:
> We may have to merge migrations because now there is more than one pending PR with a migration. This all looks good though!

What's the best way to handle that? Should I go ahead and merge this? Will those other branches just have to re-make their migrations?

annehaley:
Yes, you can merge yours and the other branch will need to either remake that migration or perform a `mergemigrations` action. The former is preferred because the latter makes a third migration file.

PR #148, opened 2022-04-27
Add a failure case to job spawning

### LINKED ISSUE #147: Jobs should have a failed status

annehaley:
Resolves #147


atlascope/core/job_types/average_color.py
```diff
@@ -17,19 +17,23 @@ def run(job_id: str, original_dataset_id: str):
...
+            )
         )
-    )
-    job.complete = True
-    job.save()
+        job.complete = True
+        job.save()
+    except Exception as e:
+        job.failure = str(e)
+        job.save()
```
waxlamp, line 46: I *think* you could factor out this `job.save()` and the one from line 36 into a `finally` clause for this try block. It is not urgent that you do that, but it would be a nice use of `finally` for the "cleanup" action of saving the Job model.

waxlamp (approved):
This looks reasonable to me.

atlascope/core/job_types/average_color.py
annehaley: Changed in [91b8359](https://github.com/atlascope/atlascope/pull/148/commits/91b835953f6a353adee000355d873ff367033a6f)

PR #159, opened 2022-05-09
Download once, populate many

### LINKED ISSUE #152: Download once, populate many times workflow
Since we are reaching the point of working with "real" data at O(GB) sizes, we will need a way to avert too much downloading of data as we work. I'd like to see a "download once, populate many times" workflow. Essentially, the dev will be responsible for downloading test data to a specific location on disk (which can be a git-ignored directory in the cloned Atlascope repository), and the populate script should have a mode where it "uploads" data from that directory to Atlascope, etc. This mode should be idempotent, which may make it a bit tricky, but it doesn't seem like anything too crazy to address.

As a first example we should use the colon image from Vandy Explorer:
- Girder item: https://styx.neurology.emory.edu/girder/#item/61fbe6214874d4ffdf1c8569
- Vandy Explorer: https://vanderbilt.digitalslidearchive.net/#!top/multichannel_view:id=61fbe6214874d4ffdf1c8569:group=0

This would be in addition to the other data import modes we have been working with. It doesn't necessarily need to be a new first-class import mode in Atlascope itself, if it's accomplishable by other means (i.e., tricks in the populate script).

mcovalt:
Related to #152

This PR uses [pooch](https://github.com/fatiando/pooch) to download files in the [`VandyImporter`](https://github.com/atlascope/atlascope/blob/cached-dev-files/atlascope/core/importers/vandy.py). The folder containing cached downloads is persisted via our Docker Compose configuration.

### DIFF
--- a/atlascope/core/importers/vandy.py
+++ b/atlascope/core/importers/vandy.py
@@ -1,6 +1,6 @@
-import io
 import os
 
+import pooch
 import requests
 
 from .base_importer import AtlascopeImporter
@@ -10,6 +10,7 @@ class VandyImporter(AtlascopeImporter):
     def perform_import(
         self,
         file_id: str,
+        sha256sum: str,
         api_key: str = None,
     ):
         api_key = api_key or os.environ.get("DJANGO_API_TOKEN")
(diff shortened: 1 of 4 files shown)


annehaley (approved):
This is cool! Pooch works really well here.

PR #169, opened 2022-05-27
Detected nucleus model

annehaley:


naglepuff:
I'm curious: if you were to run `generateTypes`, what would the `Dataset` interface look like? Would there be an optional `detected_nuclei?: DetectedNucleus` field?

naglepuff:
I'm curious: if you were to run `generateTypes`, what would the `Dataset` interface look like? Would there be an optional `detected_nuclei?: DetectedNucleus` field?

annehaley:
> I'm curious: if you were to run `generateTypes`, what would the `Dataset` interface look like? Would there be an optional `detected_nuclei?: DetectedNucleus` field?

Yes, that's how it should generate the type (though I'm not sure of the actual behavior). We can expect this field to exist on Datasets of the type `detected_nuclei`. Other types will have this field removed because it is not applicable.

atlascope/core/job_types/nucleus_detection.py
```diff
@@ -68,25 +73,60 @@ def run(job_id: str, original_dataset_id: str):
...
+                        (nucleus['Identifier.Xmax'], nucleus['Identifier.Ymin']),
+                        (nucleus['Identifier.Xmax'], nucleus['Identifier.Ymax']),
+                        (nucleus['Identifier.Xmin'], nucleus['Identifier.Ymax']),
+                        # repeat first point to create linear ring
+                        (nucleus['Identifier.Xmin'], nucleus['Identifier.Ymin']),
+                    ]
+                ),
+                **{
+                    nucleus_attribute_to_field_name(attribute): nucleus[attribute]
+                    for attribute in NUCLEUS_ATTRIBUTES
```
waxlamp: This way of expressing it is a bit odd, and took me some reading of different parts of the code to understand what's going on. At the very least, I think you should factor this expression out to before the call to `create()` and construct the dict with an expressive name, then "dump" that object here with the `**` operator.

This is purely a readability thing, but that is an important consideration in complex code such as this.

waxlamp (approved):
If you can refactor the code as I suggested, then there's no need for a second review and you can merge as soon as that's done. Thanks!

atlascope/core/job_types/nucleus_detection.py
annehaley: Ok, I changed this in [1394483](https://github.com/atlascope/atlascope/pull/169/commits/139448398be68ecada9512934a276c1ac7b11a16)

PR #170, opened 2022-06-03
Use `django-polymorphic` to better model pins

naglepuff:
This branch explores one option for better modeling pins by bringing in a Django app called `django-polymorphic`.


naglepuff:
The first batch of commits doesn't touch much client code (it regenerates types). It should capture all of the changes that we need on the back end.

naglepuff:
I've spent some time digging through the source code and github repositories for these apps, and while they are pretty small and straightforward, I'm hesitant to pull the trigger on them due to lack of activity on open issues and PRs from project maintainers. Also, `django-polymorphic` has some self-identified ["experimental" code](https://github.com/django-polymorphic/django-polymorphic/blob/bd5faf0c829f87dc6fcc42c6bfbb689d665a5bf4/polymorphic/base.py#L35), so I'm wary that this is the right approach.

I'll leave this PR as a draft to demonstrate how to use these apps to do some model inheritance in this manner, but I think it might be better to stick with vanilla Django for this functionality. Sure, we lose some of the "free stuff" that this approach would get us, like a simplified serializer and `resource_type` field, but we don't take on any risk with using third party code.

atlascope/core/models/pin.py
```diff
@@ -48,12 +44,108 @@ class Meta:
...
+        fields = '__all__'
+
+
+class DatasetPinSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = DatasetPin
+        fields = '__all__'
+
+
+class PinSerializer(PolymorphicSerializer):
```
annehaley, line 79: This is probably the most apparent difference between your PRs. I think this convoluted serializer design that comes with `django-polymorphic` is a good reason to argue in favor of vanilla model inheritance

annehaley (commented):
I can see what you mean in saying that this implementation is less ideal than your other PR. The differences are few enough that we might not notice one being more difficult than the other, but this one introduces slightly more complexity.

naglepuff:
Closing, as we've opted to take the approach in #171

atlascope/core/models/pin.py
waxlamp: I'd like to have a discussion about this when Mike is back in town. I will schedule something, because I don't think I see how this approach is convoluted, and in fact the vanilla approach seems strictly worse to me (granted, that approach doesn't bring in third party libraries, but instead it manually recreates the logic in what is likely to be both more verbose and more error prone).

PR #171, opened 2022-06-06
Explore vanilla django model inheritance

naglepuff:
This PR explores another way to use model inheritance for Pins. It differs from #170 by sticking with vanilla Django instead of bringing in third party apps `django-polymorphic` and `django-rest-polymorphic`.


atlascope/core/models/pin.py
```diff
@@ -48,12 +42,116 @@ class Meta:
...
+
+
+class DatasetPinSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = DatasetPin
+        fields = '__all__'
+
+
 class PinSerializer(serializers.ModelSerializer):
+    def to_representation(self, instance):
```
naglepuff, line 72: We now override `to_representation` in the Pin serializer. This is fine for serializing objects to JSON, and I expect this suffices for any `GET` endpoint. However, if we want to use this serializer to create objects e.g. from a `PUT` request, we'll also need to override `to_internal_value`.

`django-rest-polymorphic`'s `PolymorphicSerializer` is a subclass of DRF `Serializer` with all of this functionality build in already.

naglepuff:
I've spent some time digging through the source code and github repositories for these apps, and while they are pretty small and straightforward, I'm hesitant to pull the trigger on them due to lack of activity on open issues and PRs from project maintainers. Also, django-polymorphic has some self-identified ["experimental" code](https://github.com/django-polymorphic/django-polymorphic/blob/bd5faf0c829f87dc6fcc42c6bfbb689d665a5bf4/polymorphic/base.py#L35), so I'm wary that this is the right approach.

atlascope/core/models/pin.py
naglepuff: Although this approach may require us to do some additional work if/when we decide to use this Serializer to create new Pin objects, I think its better to not take on the technical debt of `django-polymorphic`. I've explained some of my thoughts in a comment on this PR.

annehaley (approved):
These changes look good, and they are slightly simpler than the `django-polymorphic` solution.

atlascope/core/migrations/0007_auto_20220606_1603.py
```diff
@@ -0,0 +1,72 @@
+# Generated by Django 3.2.13 on 2022-06-06 16:03
```
waxlamp, line 1: I think best practice is to rename these "auto" migration files to reflect what change they are associated with. (Do I have that right, @annehaley?)

atlascope/core/rest/endpoints/investigation_endpoints.py
```diff
@@ -25,7 +25,10 @@ class InvestigationViewSet(
     @swagger_auto_schema(responses={200: PinSerializer(many=True)})
     @action(detail=True, methods=['GET'])
     def pins(self, request, pk=None):
-        payload = PinSerializer(self.get_object().pins.all().order_by('id'), many=True).data
+        related_pins = (
+            self.get_object().pins.all().select_related('notepin', 'datasetpin').order_by('id')
+        )
+        payload = PinSerializer(related_pins, many=True).data
```
waxlamp, line 31: Why was this change necessary?

atlascope/core/models/pin.py
waxlamp: I have some serious doubts about this approach. I mentioned it in the other PR, but I'm not seeing how `django-polymorphic` adds more debt than an approach like this, where you are manually adding code clauses for each subtype. But as I said in that thread, I think this deserves a live discussion.

waxlamp (commented):
Thanks for looking into these issues, Mike!

I want to learn more so I'll schedule a time for you, me, and Anne to discuss this.

PR #174, opened 2022-06-15
Add output images to `nucleus_detection` and `nucleus_to_gland_distance` jobs

annehaley:
For the `nucleus_to_gland_distance` job, adding this computation of the output image means that we can no longer make use of previously saved calculations of nucleus locations. We need the mask of the nuclei for the visualization (which is not saved in the database), so we need to rerun the nucleus detection step within this job. This doesn't change the time delta too much.

Also, I wrote an additional step in that job to generate a random set of blobs to act as the gland mask. The one I created manually has a set size, but the gland mask needs to be the same size as the original image to use it for the output image. So for now, the resulting images (and calculated average distances) will not be deterministic unless we decide to add a seed to the random number generator.

Here are some example output images from the populate script we run:
Detected nuclei:
Detected_Nuclei

Average distance to nearest gland:
Average_Nucleus_to_Nearest_Gland_Distance


waxlamp:
I'm not super clear on why you are generating random gland masks in this job. Why can't we just use hand crafted gland maps that would be meaningful for a given image?

atlascope/core/job_types/nucleus_detection.py
```diff
@@ -63,7 +64,20 @@ def detect_nuclei(input_image):
         input_image,
     )
 
-    return additional_features.to_dict("records")
+    return (
+        additional_features.to_dict("records"),
+        np.vectorize(lambda x: 255 if x > 0 else 0)(im_nuclei_seg_mask).astype(np.uint8),
+    )
```
waxlamp, line 70: Unless there's a specific reason to return a tuple, you're usually better off using a dict instead (those are more human-readable than a bare tuple).

atlascope/core/job_types/nucleus_detection.py
```diff
@@ -63,7 +64,20 @@ def detect_nuclei(input_image):
...
+
+
+def get_result_image(input_image, nucleus_mask):
```
waxlamp, line 73: Can you include a docstring for this function? And if possible, Python type tags for the inputs and output?

IIUC, this creates a new image that overlays a nucleus mask using a green color on top of an existing image. Is that right?

atlascope/core/job_types/nucleus_detection.py
waxlamp: The reason may be the ease of destructuing the return value at the point of use; in that case, if you can add Python types to the function declaration, that would be very helpful.

atlascope/core/job_types/nucleus_to_gland_distance.py
```diff
@@ -14,70 +17,108 @@
...
-            "type": 'string',
-            "title": 'Path to gland map',
-            "default": 'atlascope/core/management/populate/inputs/gland_map.jpg',
-        }
-    },
+    "properties": {},
 }
 
 
+def generate_gland_map(input_image):
```
waxlamp, line 24: Repeating myself but I'm not clear on why this function needs to exist. As for controlling randomness, if this does need to be here, I would make the RNG seed an optional argument with a default value.

atlascope/core/job_types/nucleus_to_gland_distance.py
```diff
@@ -14,70 +17,108 @@
...
-def run(job_id: str, original_dataset_id: str, gland_map_path: str):
+def run(job_id: str, original_dataset_id: str):
+    # suppress warnings from high numbers and NaNs in computations
+    # we only need this when we are using a randomly generated gland map
+    import warnings
+
     from atlascope.core.models import Job
 
+    # TODO: remove this warning suppression
+    warnings.filterwarnings("ignore")
```
waxlamp, line 67: This seems like a bad idea. Better could be to constrain the random mask generation not to cause these warnings in the first place (if possible).

atlascope/core/job_types/nucleus_to_gland_distance.py
```diff
@@ -14,70 +17,108 @@
...
     job = Job.objects.get(id=job_id)
 
     try:
-        # If a previous job has nucleus detections, use that information
-        nucleus_detections = Job.objects.filter(
-            original_dataset=original_dataset,
-            job_type='nucleus_detection',
+        # We cannot use previous nucleus detection results because we do not
+        # save the nucleus mask to the database
+        # (we need the nucleus mask for the output image)
```
waxlamp, line 75: Could we save it to the DB? What's blocking us from doing so currently?

(Recomputing it like this is ok too, but it represents a bit of technical debt, so it's worthwhile to have the discussion.)

atlascope/core/job_types/nucleus_to_gland_distance.py
```diff
@@ -14,70 +17,108 @@
...
+        for nucleus_centroid in nucleus_centroids:
+            minimum = None
+            closest = None
+            for gland_centroid in gland_centroids:
+                distance = math.dist(nucleus_centroid, gland_centroid)
+                if not minimum or distance < minimum:
+                    minimum = distance
+                    closest = gland_centroid
+            distances.append(minimum)
+            distances_mask_draw.line([nucleus_centroid, closest], width=1)
```
waxlamp, line 105: This is *O(mn)* so if we find it's too slow for our examples, we can apply some acceleration structures here. A simple one is to split the image up into a grid, then only check nuclear-gland distances in the grid region occupied by the nucleus, and the eight neighboring regions. (This is a bit tricky, since if e.g. there are no glands in the nine regions, then you'd have to expand the search until you find one; this complexity is why we should only turn to this approach if we run into performance issues here.)

atlascope/core/management/populate/jobs.json
```diff
@@ -24,9 +24,6 @@
         "job_type": "nucleus_to_gland_distance",
         "id": 4003,
         "investigation": "Analytics playground",
-        "original_dataset": "HTA9_1_BA_F_ROI04",
-        "additional_inputs": {
-            "gland_map_path": "atlascope/core/management/populate/inputs/gland_map.jpg"
-        }
+        "original_dataset": "HTA9_1_BA_F_ROI04"
```
waxlamp, line 27: Why dump the gland map input? (I suppose this is connected to why we're introducing random gland map generation in this PR.)

annehaley:
Closing for now in favor of client-side visualization

PR #177, opened 2022-06-21
Smooth out pin transition during zoom in/out

### LINKED ISSUE #175: Smooth out Pin transition when zooming
If a user zooms beyond the maximum or minimum zoom level for a pin, that pin will pop out of existence. For a better user experience, it would be nice if there was a smoother transition. One idea is the change the radius or opacity of the pin annotations as the user approaches their zoom range.

naglepuff:
This PR attempts to enhance user experience by introducing simple transitions as a user zooms through a map, causing pins to appear/disappear based on their minimum and maximum zoom levels.

While when approaching a pin by zooming, the pin will appear to grow until the user hits the minimum zoom level for that pin. When zooming beyond a pin's maximum zoom level, its opacity will decrease from maximum to 0. I believe this nicely simulates zooming "past" the scale where the pin would be useful to display, but I'm open to other ideas for transitions.

Resolves #175

### DIFF
--- a/web/src/utilities/utiltyFunctions.ts
+++ b/web/src/utilities/utiltyFunctions.ts
@@ -17,3 +17,27 @@ export function postGisToPoint(location: string): Point | undefined {
   }
   return { x: xCoord, y: yCoord };
 }
+
+export function radiusForZoomLevel(currentZoom: number, minZoom: number): number {
+  const maxRadius = 10;
+  if (currentZoom >= minZoom) {
+    return maxRadius;
+  }
+  const diff = minZoom - currentZoom;
+  if (diff > 2) {
+    return 0;
+  }
+  return maxRadius - (5 * diff);
+}
+
+export function opacityForZoomLevel(currentZoom: number, maxZoom: number): number {
+  const maxOpacity = 0.8;
+  if (currentZoom <= maxZoom) {
+    return maxOpacity;
+  }
+  const diff = currentZoom - maxZoom;
+  if (diff > 2) {
+    return 0;
+  }
+  return maxOpacity - (0.4 * diff);
+}
--- a/web/src/utilities/utiltyFunctions.ts
+++ b/web/src/utilities/utiltyFunctions.ts
@@ -18,16 +18,24 @@ export function postGisToPoint(location: string): Point | undefined {
   return { x: xCoord, y: yCoord };
 }
 
-export function radiusForZoomLevel(currentZoom: number, minZoom: number): number {
-  const maxRadius = 10;
-  if (currentZoom >= minZoom) {
-    return maxRadius;
+export function radiusForZoomLevel(currentZoom: number, minZoom: number, maxZoom: number): number {
+  const inRangeRadius = 10;
+  let diff = 0;
+  if (currentZoom >= minZoom && currentZoom <= maxZoom) {
+    return inRangeRadius;
   }
-  const diff = minZoom - currentZoom;
+  if (currentZoom < minZoom) {
+    diff = minZoom - currentZoom;
+    if (diff > 2) {
+      return 0;
+    }
+    return inRangeRadius - (5 * diff);
+  }
+  diff = currentZoom - maxZoom;
   if (diff > 2) {
-    return 0;
+    return inRangeRadius;
   }
-  return maxRadius - (5 * diff);
+  return inRangeRadius + (8 * diff);
 }
 
 export function opacityForZoomLevel(currentZoom: number, maxZoom: number): number {
--- a/web/src/views/InvestigationDetail.vue
+++ b/web/src/views/InvestigationDetail.vue
@@ -259,9 +259,10 @@ export default defineComponent({
       if (!featureLayer || !pinFeature || !map.value) { return; }
       pinFeature.getData().forEach((pin: object) => {
         if (!pinFeature) return;
-        const { x, y } = postGisToPoint((pin as Pin).location) || { x: 0, y: 0 };
+        const pinObject = pin as Pin;
+        const { x, y } = postGisToPoint(pinObject.location) || { x: 0, y: 0 };
         const newScreenCoords = pinFeature.featureGcsToDisplay(x, y);
-        const note = pinNotes.value.find((pinNote) => pinNote.id === (pin as Pin).id);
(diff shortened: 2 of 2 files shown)


mcovalt (approved):
This works really well. Should we do a similar transition for the pins' notes? Also, I noticed that notes for minimum-zoom-level-pins get strangely offset when zooming.

naglepuff:
I'm inclined to punt on the pin notes, so we can get this PR merged faster. I've opened #187 to track this idea. Would you mind sharing a screenshot of the strange offset you're seeing? I'm having some difficulty reproducing that issue.

mcovalt:
> I'm inclined to punt on the pin notes, so we can get this PR merged faster.

Totally agree. I've approved and think it's ready to merge as-is. Thanks for opening that issue!

> Would you mind sharing a screenshot of the strange offset you're seeing? I'm having some difficulty reproducing that issue.

Sure, I'll get that posted here in an hour or so.

naglepuff:
Thanks. I'll go ahead and merge this. I'll keep an eye out for that screenshot. Feel free to open an issue for that as well.

web/src/utilities/utiltyFunctions.ts
```diff
@@ -17,3 +17,35 @@ export function postGisToPoint(location: string): Point | undefined {
...
+  let diff = 0;
+  if (currentZoom >= minZoom && currentZoom <= maxZoom) {
+    return inRangeRadius;
+  }
+  if (currentZoom < minZoom) {
+    diff = minZoom - currentZoom;
+    if (diff > 2) {
+      return 0;
+    }
+    return inRangeRadius - (5 * diff);
```
waxlamp, line 32:
```suggestion
    // While zooming out, ramp down radius from full size to zero.
    return Math.max(inRangeRadius - 5 * (minZoom - currentZoom), 0.0);
```

This might seem harder to read, but I think it's actually better. It's a more succinct description of what's going on, involving fewer `if`s and intermediate computed values, and the comment helps to guide the reader to exactly what's going on.

mcovalt:
> Would you mind sharing a screenshot of the strange offset you're seeing? I'm having some difficulty reproducing that issue.

@naglepuff Well, I can't reproduce either anymore 🤔. I'll keep an eye out if I see it again, but I say "issue solved".

web/src/utilities/utiltyFunctions.ts
waxlamp: And actually, an even better approach might be to define an `interpolate()` function somewhere (or use one from the standard library) and use it here directly.

web/src/utilities/utiltyFunctions.ts
```diff
@@ -17,3 +17,35 @@ export function postGisToPoint(location: string): Point | undefined {
   }
   return { x: xCoord, y: yCoord };
 }
+
+export function radiusForZoomLevel(currentZoom: number, minZoom: number, maxZoom: number): number {
+  const inRangeRadius = 10;
+  let diff = 0;
+  if (currentZoom >= minZoom && currentZoom <= maxZoom) {
```
waxlamp, line 24: I would restructure the sequence of `if` statements to be "if-else if-else" instead--this will improve readability and it will more strongly express the disjoint nature of the test. I am thinking something like `if (currentZoom < minZoom)`, then `else if (currentZoom > maxZoom)`, then `else` (to handle the "in range" portion of the test).

web/src/utilities/utiltyFunctions.ts
```diff
@@ -17,3 +17,35 @@ export function postGisToPoint(location: string): Point | undefined {
...
+    if (diff > 2) {
+      return 0;
+    }
+    return inRangeRadius - (5 * diff);
+  }
+  diff = currentZoom - maxZoom;
+  if (diff > 2) {
+    return inRangeRadius;
+  }
+  return inRangeRadius + (8 * diff);
```
waxlamp, line 38: Use a similar technique here as above (using `Math.min()` this time, with an appropriate ceiling value).

web/src/utilities/utiltyFunctions.ts
```diff
@@ -17,3 +17,35 @@ export function postGisToPoint(location: string): Point | undefined {
...
+export function opacityForZoomLevel(currentZoom: number, maxZoom: number): number {
+  const maxOpacity = 0.8;
+  if (currentZoom <= maxZoom) {
+    return maxOpacity;
+  }
+  const diff = currentZoom - maxZoom;
+  if (diff > 2) {
+    return 0;
+  }
+  return maxOpacity - (0.4 * diff);
```
waxlamp, line 50: Use similar technique as above.

web/src/views/InvestigationDetail.vue
```diff
@@ -254,9 +259,10 @@ export default defineComponent({
       if (!featureLayer || !pinFeature || !map.value) { return; }
       pinFeature.getData().forEach((pin: object) => {
         if (!pinFeature) return;
-        const { x, y } = postGisToPoint((pin as Pin).location) || { x: 0, y: 0 };
+        const pinObject = pin as Pin;
```
waxlamp, line 262: There is an improvement to this style of "casting" (which is generally unsafe in this context). You should use a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) here to ensure the values are in fact of the right type (if, instead, it's not possible to type `pinFeature.getData()` as `Pin[]`). Let me know if you want to discuss this further.

waxlamp (changes requested):
On further reflection, I think a dedicated `interpolate()` function would be useful for factoring out much of the logic in the radius ramping functions.

This looks great overall. Thanks!

web/src/utilities/utiltyFunctions.ts
naglepuff: `eslint` actually gets mad if I turn these into `if/else` statements, since I'm returning a value within the `if` block.

waxlamp: In that case, you're still better off doing something like setting up a `radius` variable, setting it in the cascade of if-else if-else tests, and returning a single value at the end.

Generally speaking, you should prioritize readability and idiomatic usage (which this change would be), and satisfy the linter under those constraints.

PR #180, opened 2022-06-23
Anatomy datasets

annehaley:
This PR pulls the content of twelve tiled tiff images from our shared Google Drive. When the populate script is run, pooch will fetch and store them. This happens within a new importer called the `GoogleDriveImporter`.

My experience has been that Google allows ten requests, though, before you exceed the limit and you have to try again later. I added some exception handling for this, so on the first attempt, you'll get 10 of them in your "Anatomy" investigation. If you wait and run the populate script again the next day (or just some time later, I'm not sure about Google's API limit window), then the first ten will be still saved by pooch and the other two downloads will start. Then you'll have all twelve in the "Anatomy" investigation. 

image

### DIFF
--- a/atlascope/core/importers/__init__.py
+++ b/atlascope/core/importers/__init__.py
@@ -1,6 +1,6 @@
+from .gdrive import GoogleDriveImporter
 from .upload import UploadImporter
 from .vandy import VandyImporter
-from .gdrive import GoogleDriveImporter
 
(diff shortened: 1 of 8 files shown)


mcovalt (approved):
Awesome! If our CI pings Google too much, then we can persist the `/root/.cache/pooch/` directory in GitHub Actions.

PR #185, opened 2022-06-24
Add select/deselect all button to Pin List

### LINKED ISSUE #184: Add show/hide all button in Pin List

naglepuff:
Resolves #184 

This PR adds a checkbox button at the top of the pin list that allows selecting/deselecting all pins at once. 
image


mcovalt (commented):
It works!

This includes some code changes from #177. I'm going to wait until that PR is merged and this branch merges in those upstream changes. After that, the #177 code ought to disappear and then I'll approve.

mcovalt (approved):

waxlamp (commented):
Looks good overall. I think a slightly cleaner approach would make use of the checkbox's `v-model` property, as in [this example](https://stackoverflow.com/a/57752579/1886928) from Stack Overflow. I'd like you to investigate whether this approach would work here, and use it if so (I find the connection to `v-model` using Vue idioms is a better approach than checking that the number of checked items is equal to the full length of options).

naglepuff:
> Looks good overall. I think a slightly cleaner approach would make use of the checkbox's `v-model` property, as in [this example](https://stackoverflow.com/a/57752579/1886928) from Stack Overflow. I'd like you to investigate whether this approach would work here, and use it if so (I find the connection to `v-model` using Vue idioms is a better approach than checking that the number of checked items is equal to the full length of options).

I'm not sure I agree here. In the example you've shared, someone has created a computed property with both a getter and setter. After re-reading the Vue [documentation for computed properties](https://vuejs.org/guide/essentials/computed.html#best-practices), using a setter with a computed property seems like it should be used in rare circumstances. Since the state of the new checkbox can be totally derived from existing state, I don't think we need to follow the example you shared. Let me know if you think I'm missing something.

waxlamp:
> > Looks good overall. I think a slightly cleaner approach would make use of the checkbox's `v-model` property, as in [this example](https://stackoverflow.com/a/57752579/1886928) from Stack Overflow. I'd like you to investigate whether this approach would work here, and use it if so (I find the connection to `v-model` using Vue idioms is a better approach than checking that the number of checked items is equal to the full length of options).
> 
> I'm not sure I agree here. In the example you've shared, someone has created a computed property with both a getter and setter. After re-reading the Vue [documentation for computed properties](https://vuejs.org/guide/essentials/computed.html#best-practices), using a setter with a computed property seems like it should be used in rare circumstances. Since the state of the new checkbox can be totally derived from existing state, I don't think we need to follow the example you shared. Let me know if you think I'm missing something.

This seems like a case where you are doing non-standard things to avoid a "rare circumstance" (as called out by the docs) from manifesting.

If you look at your setup, you'll find both a getter (in, e.g., the `input-value` prop of the `v-checkbox`) and a setter (in your `toggleDisplayAll()` function) all linked to that checkbox. On the other hand, if you simply track the checkbox's state via its v-model prop, and accept that, although the state of that checkbox is a function of other inputs on the page, that it can also *influence* that state, then a computed setter does seem to be appropriate here. That way, you're leaning into the fact that there's a two-way flow of information through this element, and embracing standard Vue mechanisms to implement it.

I agree that it's a bit fuzzy, so I'd like to meet with you to pair program/explore this. I'm insisting on this out of a desire to better understand Vue best practices myself, and transmit that to the teams I'm a part of.

waxlamp:
We could also just ask @subdavis what he thinks 😎

naglepuff:
> I agree that it's a bit fuzzy, so I'd like to meet with you to pair program/explore this. I'm insisting on this out of a desire to better understand Vue best practices myself, and transmit that to the teams I'm a part of.

We met up to do some pair programming and compare the existing changes with using `v-model`. To do this, we created a new computed property with both a `get` and `set`. We transplanted the `toggleDisplayAll` function into the setter of the new computed property. While this did improve the readability of the code by putting the logic related to the checkbox in one place (a new computed property), it opened a different can of worms.

Our new computed property, `allSelected`, is derived from the properties `selectedPins` and `pins`. If the length of both of these arrays of `Pins` is equal, then all pins are selected. So, if we use the computed setter like `allSelected.value = true`, we know what the side effects of this action will be. We must add all `Pin` objects in `pins` to `selectedPins`. However, what if someone down the line writes `allSelected = false` somewhere? What is the side effect of setting the computed property like this? In the case of the select/deselect all checkbox, we know that we set `selectedPins` to an empty array (deselect all). But if a statement like this is written somewhere else, what should the setter do? It can't possibly know how to update `selectedPins`. In a case like this, what should the setter do? Nothing? I'd rather not have a nonsensical setter on a computed property, as it could be misleading to future observers of the code base.

At this point I'd argue in favor of handling the `change` event of the checkbox instead of creating a new computed property with a dubious setter in order to use `v-model`.

waxlamp:
> But if a statement like this is written somewhere else, what should the setter do?

I don't remember where we landed on this in our live discussion, but my gut tells me that if someone wrote `allSelected.value = false` somewhere, we should deactivate all pins. That's what happens in GMail, for example.

PR #186, opened 2022-06-27
List in-bounds pins first

### LINKED ISSUE #181: Sort pins by visiblity in sidebar
We could enhance the pin list in the sidebar by prioritizing pins that are currently visible to the user based on current image bounds and zoom level. Related to #176

naglepuff:
Resolves #181 

This PR adds the map bounds to the store and a new computed property to the pin list component. These additions are used to sort the pins that are within the map bounds to the top of the list in the sidebar. As a user zooms/pans around the maps, the pins that become visible jump up the list.

To do: 

- [x] Factor zoom level into sorting function


web/src/components/PinList.vue
```diff
@@ -56,11 +56,40 @@ import {
...
 
 export default defineComponent({
   setup() {
     const pins: Ref<Pin[]> = computed(() => store.state.currentPins);
+    const bounds = computed(() => store.state.currentBounds);
+    const zoom = computed(() => store.state.zoomLevel);
     const rootDataset = computed(() => store.state.rootDataset);
+
+    function isInBounds(pin: Pin): boolean {
+      const point: Point = postGisToPoint(pin.location) || { x: 0, y: 0 };
```
waxlamp: Looking at the source for `postGisToPoint()`, it doesn't look like it truly returns `Point | undefined`; all the error conditions lead to throwing an exception.

I bring it up because this code doesn't look correct to me--if `postGisToPoint()` were to fail, you wouldn't want to continue from this point using (0, 0). I think the practical effect is to crash the program, but this code falsely implies that it has a chance of continuing from this point with bad data. I think it would be best to fix up the types for that function, and remove the `||` part from this line.

web/src/components/PinList.vue
```diff
@@ -56,11 +56,40 @@ import {
...
+      const containedInMapBounds = (point.x < bounds.value.right
+        && point.x > bounds.value.left
+        && point.y > bounds.value.top
+        && point.y < bounds.value.bottom);
```
waxlamp, line 73: Is there a GeoJS function that does this check for you? It seems like the kind of thing that would exist as a utility function, but maybe not.

web/src/components/PinList.vue
```diff
@@ -56,11 +56,40 @@ import {
...
+      const pinMaxZoom = pin.maximum_zoom || 40;
+      const pinMinZoom = pin.minimum_zoom || 0;
+      const visibleAtCurrentZoom = (
+        pinMinZoom - 2 <= zoom.value
+        && pinMaxZoom + 2 >= zoom.value
```
waxlamp: For readability, you may want to invert this test:

```suggestion
        && zoom.value <= pinMaxZoom + 2
```

waxlamp (approved):
Looks good! If you accept my suggestions, no need for another review. Otherwise, you can make changes and request another review.

web/src/components/PinList.vue
naglepuff: 4fc958a

web/src/components/PinList.vue
naglepuff: I've asked the geojs experts, but while waiting for a response the best thing I could find on my own was the `boxSearch` [function](https://opengeoscience.github.io/geojs/apidocs/geo.feature.html#boxSearch). We could call this function, which is a member of `pinFeature` to get a list of all pins within a given box. We could then compare the pin in question with the list to see if its within the box. There doesn't seem to be a straightforward function on `geo.map` that helps answer the question: "is this point in bounds?"

I think the current method of sorting the list of pins based on current bounds is more straightforward than getting a list of currently visible pins and comparing it to the list of all pins.

PR #190, opened 2022-07-01
Tour UI

marySalvi:
Adds UI for Tour Model. 
User can step through Tour Waypoints or pick them from the list individually.

https://user-images.githubusercontent.com/40494088/179044769-605c4122-3d04-436b-87ad-3da36e9e2abc.mp4

### DIFF
--- a/atlascope/core/management/populate/tours.json
+++ b/atlascope/core/management/populate/tours.json
@@ -1,5 +1,5 @@
 [{
-    "investigation": "Pin playground",
+    "investigation": "Tour playground",
     "id": 6000,
     "name": "Tour Demo 1",
     "waypoints": [
@@ -30,7 +30,7 @@
     ]
 },
 {
-    "investigation": "Pin playground",
+    "investigation": "Tour playground",
     "id": 6001,
     "name": "Tour Demo 2",
     "waypoints": [
--- a/atlascope/core/management/populate/tours.json
+++ b/atlascope/core/management/populate/tours.json
@@ -1,30 +1,61 @@
 [{
     "investigation": "Pin playground",
     "id": 6000,
-    "name": "Tour Demo",
+    "name": "Tour Demo 1",
     "waypoints": [
         {
-            "location": [0.0, 0.0],
+            "location": [1200.0, 3000.0],
             "zoom": 1
         },
         {
-            "location": [1.0, 1.0],
+            "location": [2200.0, 2200.0],
             "zoom": 2
         },
         {
-            "location": [2.0, 2.0],
+            "location": [1100.0, 1100.0],
             "zoom": 3
         },
         {
-            "location": [3.0, 3.0],
+            "location": [3300.0, 1100.0],
             "zoom": 1
         },
         {
-            "location": [4.0, 4.0],
+            "location": [3300.0, 3300.0],
             "zoom": 2
         },
         {
-            "location": [5.0, 5.0],
+            "location": [4400.0, 4400.0],
+            "zoom": 3
+        }
+    ]
+},
+{
+    "investigation": "Pin playground",
+    "id": 6001,
+    "name": "Tour Demo 2",
+    "waypoints": [
+        {
+            "location": [1100.0, 1100.0],
+            "zoom": 1
+        },
+        {
+            "location": [1200.0, 3000.0],
+            "zoom": 2
+        },
+        {
+            "location": [2200.0, 2200.0],
+            "zoom": 3
+        },
+        {
+            "location": [3300.0, 3300.0],
+            "zoom": 1
+        },
+        {
+            "location": [4400.0, 4400.0],
+            "zoom": 2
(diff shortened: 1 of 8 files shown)


atlascope/core/management/populate/tours.json
```diff
@@ -1,30 +1,61 @@
 [{
     "investigation": "Pin playground",
```
naglepuff: Is there a particular reason you had in mind for adding the tours to this investigation? It might be better to create a new investigation in the populate data for tours.

web/src/utilities/useGeoJSLayer.ts
```diff
@@ -1,5 +1,5 @@
-import { ref } from '@vue/composition-api';
-import useGeoJSFeature from './useGeoJSFeature';
+import { ref } from "@vue/composition-api";
```
naglepuff: Looks like a few of these files got reformatted, and the client lint tests are no longer passing.

web/src/components/TourList.vue
```diff
@@ -0,0 +1,93 @@
...
+    item-key="id"
+  >
+    <template v-slot:expanded-item="{ headers, item }">
+      <td :colspan="headers.length" class="white">
+        <v-stepper v-model="currentStep" vertical>
+          <template>
+            <div class="d-flex justify-space-between mx-5">
+              <v-btn
+                color="primary"
+                :disabled="currentStep == 0 || currentStep == 1"
```
naglepuff: I think its better practice to use `===` for comparison here

web/src/components/TourList.vue
```diff
@@ -0,0 +1,93 @@
...
+                @click="previousStep()"
+              >
+                Previous
+              </v-btn>
+              <v-btn
+                color="primary"
+                :disabled="currentStep == item.waypoints.length"
```
naglepuff: Same here. I'd change `==` to `===`

web/src/store/index.ts
```diff
@@ -39,6 +39,9 @@ export interface State {
     selectionMode: boolean;
     subimageSelection: number[] | null;
     jobTypes: JobDetail[];
+    tours: Tour[];
+    selectedTour: Tour[];
```
naglepuff: There seems to be a mismatch between the name `selectedTour` and its type of `Tour[]`. I'd expect it to just be one tour that's selected.

naglepuff (changes requested):
Really cool to see this in action. I noticed a lot of style changes here that conflict with the existing style.

web/src/utilities/useGeoJSLayer.ts
marySalvi: I had vetur disabled and I think it might have messed with eslint as a result. Should be fixed in f369babfc6c4edb62d2b93969ed102ea42c083cf

annehaley:
That demo video is awesome!

atlascope/core/management/populate/tours.json
marySalvi: Originally wanted to show juxtaposition with Pins but I agree it makes sense for it to be on its own. Changed in 5f2b6f38fa9494bf726d0a1158d509a77aae0c3f

web/src/components/TourList.vue
marySalvi: Changed in daf78f249df149db85dfc37b41c4e39340088e96

web/src/components/TourList.vue
marySalvi: Changed in daf78f249df149db85dfc37b41c4e39340088e96

web/src/store/index.ts
marySalvi: Changed in daf78f249df149db85dfc37b41c4e39340088e96

naglepuff: Thanks. I see now that the `expanded` prop for the vuetify grid expects an array. I still think that `selectedTour` in the store should just be a `Tour` object, in case other components end up caring about it.

naglepuff (approved):
Looks good to me! I think its at a good point for merging. A couple of follow up ideas I have are:

- Ability to turn off the animations
- Ability to "clear" the currently selected tour (remove the geojs feature when no tour is expanded)

web/src/store/index.ts
waxlamp: Agreed. If some API requires a list of objects, and you want to maintain a single one, that sounds like a nice application for a computed property that yields a singleton list.

waxlamp:
This is great, @marySalvi!

My one suggestion is to change the default method of interpolating between waypoints. This seems to be using a straight linear interpolation (move from A to B at a constant speed, with "infinite" acceleration to full speed or 0 at the beginning and end, respectively), where it might feel more natural to use a form of "easing" (that is, a velocity curve that accelerates and decelerates smoothly at the ends).

I'm not sure how simple this is in GeoJS, but we should dig into it.

PR #191, opened 2022-07-11
Sort by structural similarity

### LINKED ISSUE #155: Set up a "nucleus similarity search" job type
The intent here is to select one of the nuclei and find other nuclei that are "similar" in some structural sense. The appropriate skimage function might be [structural_similarity](https://scikit-image.org/docs/stable/api/skimage.metrics.html#skimage.metrics.structural_similarity).

mcovalt:
Closes #155

This PR adds:
- a new endpoint `/nucleus-detections/{id}/similar` that returns a list of similar nuclei
- the field `fingerprint` to `DetectedNucleus`
- the "dissimilarity" measure to the response of `/nucleus-detections/{id}/similar`
-  a new endpoint `/nucleus-detections/{id}/fingerprint` that returns an image representation of a fingerprint


.github/workflows/ci.yml
```diff
@@ -17,12 +17,7 @@ jobs:
         test-type: [lint, check-migrations, test]
     services:
       postgres:
-        # Heroku uses PostGIS 2.5.5 and PostgreSQL 13.5
-        # This is technically not a valid combination
-        # according to PostGIS and is therefore not
-        # not an image they offer. We use PostgreSQL 12.x
-        # and PostGIS 2.5.x instead.
-        image: postgis/postgis:12-2.5
+        image: postgis/postgis:14-3.2
```
annehaley, line 20: A note for others: when I tried out this branch, postgis would not start with the following error. 

```
2022-08-01 19:30:07.896 UTC [1] DETAIL:  The data directory was initialized by PostgreSQL version 12, which is not compatible with this version 14.4 (Debian 14.4-1.pgdg110+1).
```

I performed a `docker-compose down` to clear the db container completely. Then You can start anew with `docker-compose up` and `python manage.py migrate && python manage.py populate` (within the django container as always)

atlascope/core/models/detected_nucleus.py
```diff
@@ -112,12 +118,29 @@ class DetectedNucleus(models.Model):
...
+    """Return a queryset of nuclei ordered by similarity."""
+    if queryset is None:
+        queryset = DetectedNucleus.objects.all()
+    return queryset.annotate(
+        dissimilarity=CubeDistance('fingerprint', nucleus.fingerprint)
+    ).order_by('dissimilarity')
+
+
 class DetectedNucleusSerializer(serializers.ModelSerializer):
+    fingerprint = serializers.ListField(
```
annehaley: Will we use this field for client side visualization? If not, it is a large amount of data to transfer and we'll need a reason to include it in the serializer

atlascope/core/models/detected_nucleus.py
```diff
@@ -112,12 +118,29 @@ class DetectedNucleus(models.Model):
     )
 
 
+def similar_nuclei(nucleus, queryset=None):
+    """Return a queryset of nuclei ordered by similarity."""
+    if queryset is None:
+        queryset = DetectedNucleus.objects.all()
+    return queryset.annotate(
+        dissimilarity=CubeDistance('fingerprint', nucleus.fingerprint)
+    ).order_by('dissimilarity')
```
annehaley, line 127: Does this order from least dissimilar to most dissimilar?

atlascope/core/fields/cube.py
```diff
@@ -0,0 +1,56 @@
...
+
+from django.db.models import Field, FloatField, Func
+from psycopg2.extensions import AsIs
+
+from atlascope.core.utils import register_adapter
+
+__all__ = ['Cube', 'CubeDistance', 'CubeField']
+
+
+class Cube(UserList):
```
annehaley, line 13: Could you explain the purpose of this addition? It seems like this is the data structure we use to store a fingerprint, but why is it a Cube?

atlascope/core/models/detected_nucleus.py
mcovalt: I had planned on using this for client side visualization, but on second thought I think not. I'll drop this serializer in preference for a `/nucleus-detections/{id}/fingerprint` endpoint that encodes this fingerprint into an image server-side.

atlascope/core/models/detected_nucleus.py
mcovalt: Yes. This orders in ascending order the "dissimilarity" measure. An image is identical if `dissimilarity == 0`. Higher values of "dissimilarity" mean the images differ more.

atlascope/core/fields/cube.py
mcovalt: Great question. We define `Cube` so that [psycopg2](https://www.psycopg.org/docs/) can convert it to appropriate SQL. [This function](https://github.com/atlascope/atlascope/pull/191/files#diff-db5454fe2adc4a5225b50335b0c1f73e85f54a1383549418e98e93ac8caff21cR17-R20) defines that behavior. The SQL translation would be incorrect for just a plain list.

atlascope/core/models/detected_nucleus.py
mcovalt: Fixed in 83fd6218d62fa3db4967f7c51ef576e1dc4ae5ee and 63666378135dae79335ca55d34396624f652b686.

atlascope/core/rest/endpoints/detected_nucleus_endpoints.py
```diff
@@ -18,3 +25,35 @@ def get_queryset(self):
...
+    @action(detail=True, methods=['get'])
+    def similar(self, request, pk=None):
+        nucleus = self.get_object()
+        queryset = similar_nuclei(nucleus)
+        page = self.paginate_queryset(queryset)
+        if page is not None:
+            serializer = SimilarNucleusSerializer(page, many=True)
+            return self.get_paginated_response(serializer.data)
+        serializer = SimilarNucleusSerializer(queryset, many=True)
+        return Response(serializer.data)
```
annehaley, line 38: does this mean the response may or may not be paginated? Does this change the shape of the response (e.g . keys "next" and "previous")?

annehaley (approved):
Do you have a vision yet for how the web client will use this information in a visualization? I'd be glad to help there.

atlascope/core/rest/endpoints/detected_nucleus_endpoints.py
mcovalt: This is taken directly from DRF's [`ListModelMixin`](https://github.com/encode/django-rest-framework/blob/71e6c30034a1dd35a39ca74f86c371713e762c79/rest_framework/mixins.py#L33-L46). Ideally it should act as a paginated response in the same way other `list`-based DRF views would.

atlascope/core/fields/cube.py
```diff
@@ -0,0 +1,56 @@
...
+    data: Iterable[float]
+
+
+@register_adapter(Cube)
+def adapt_cube(cube: Iterable[float]) -> str:
+    quoted_vector = [str(x) for x in cube]
+    return AsIs("cube(array[%s]::float[])" % ",".join(quoted_vector))
```
waxlamp, line 20: I would have probably argued for using an f-string here instead of the `%`-based interpolation operator.

atlascope/core/fields/cube.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def parse_cube(cube_string: str) -> Cube:
+    quoted_vector = re.findall(r'-?\d+(?:\.\d+)?(?:[eE]-?\d+)?', cube_string)
```
waxlamp, line 24: Oof. Is there not a more standard way to detect floating point numbers in strings?

atlascope/core/fields/cube.py
```diff
@@ -0,0 +1,56 @@
...
+    data: Iterable[float]
+
+
+@register_adapter(Cube)
+def adapt_cube(cube: Iterable[float]) -> str:
+    quoted_vector = [str(x) for x in cube]
+    return AsIs("cube(array[%s]::float[])" % ",".join(quoted_vector))
+
+
+def parse_cube(cube_string: str) -> Cube:
```
waxlamp, line 23: This function seems like it should be a class method of `CubeField`.

atlascope/core/fields/cube.py
```diff
@@ -0,0 +1,56 @@
...
+            return value
+        return parse_cube(value)
+
+    def to_python(self, value):
+        if isinstance(value, Cube) or value is None:
+            return value
+        return parse_cube(value)
+
+
+class CubeDistance(Func):
```
waxlamp, line 46: Out of curiosity: what's an example of how you would end up using this?

waxlamp (commented):
Very cool stuff. None of my comments in this PR are urgent to address, but please do read over and think about them.

atlascope/core/fields/cube.py
mcovalt: The `CubeDistance` is how we're finding similar nuclei. See [this line](https://github.com/atlascope/atlascope/pull/191/files#diff-b6cbc5f32a7e320716eb17b588e7207a7f5a071645c18ec9f2696051c9e5ff02R126). Also, see the [relevant PostgreSQL docs](https://www.postgresql.org/docs/current/cube.html#CUBE-OPERATORS-TABLE). 

Two nuclei that are similar will have a small Euclidean distance between them since each nucleus is indexed as a 100D vector with unit length. The GiST index on the `CubeField` clusters nearby nuclei on a single page. So, ordering all nuclei by their distance to a target nucleus and picking just the top few can result in an $\mathcal{O}(1)$ search.

atlascope/core/fields/cube.py
mcovalt: The string is represented as `"(x_1,x_2,x_3,...,x_n)"`, so something along the lines of...
```python
floats: list[float] = [float(sub_string) for sub_string in cube_string[1:-2].split(",")]
```
...might be preferable to a regex.

waxlamp: What you have is probably fine, though a comment explaining the format of the string would be good information to have there.

PR #192, opened 2022-07-14
Follow up Pin transition changes

### LINKED ISSUE #188: Follow-up on Pin Zoom PR
On further reflection, I think a dedicated `interpolate()` function would be useful for factoring out much of the logic in the radius ramping functions.

_Originally posted by @waxlamp in https://github.com/atlascope/atlascope/pull/177#pullrequestreview-1022217688_

See review comments on PR #177 for more details.

naglepuff:
This PR addresses issues from PR #177 

Resolves #188

### DIFF
--- a/web/src/views/InvestigationDetail.vue
+++ b/web/src/views/InvestigationDetail.vue
@@ -264,35 +264,49 @@ export default defineComponent({
       pinFeature = undefined;
     }
 
+    function isPin(obj: object): obj is Pin {
+      return (
+        'id' in obj
+        && 'location' in obj
+        && 'minimum_zoom' in obj
+        && 'maximum_zoom' in obj
+      );
+    }
+
     function movePinNoteCards() {
-      if (!featureLayer || !pinFeature || !map.value) { return; }
+      if (!featureLayer || !pinFeature || !map.value) {
+        return;
+      }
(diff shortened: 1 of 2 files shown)


web/src/views/InvestigationDetail.vue
```diff
@@ -255,35 +255,45 @@ export default defineComponent({
...
 
+    function isPin(obj: object): obj is Pin {
+      return (
+        'id' in obj
+        && 'location' in obj
+        && 'minimum_zoom' in obj
+        && 'maximum_zoom' in obj
+      );
+    }
+
```
annehaley, line 275: Could you explain the purpose of this function? Why doesn't typescript take care of this for you?

annehaley (approved):

web/src/views/InvestigationDetail.vue
naglepuff: Typescript can't tell at runtime if an object implements an interface (https://stackoverflow.com/questions/31748277/instanceof-check-on-interface).

This function allows for that to happen. Its return type is a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).

waxlamp (commented):
LGTM.

PR #193, opened 2022-07-19
Job result Visualizations

annehaley:
This PR introduces client-side visualization of job results. On the backend, we detect nuclei and glands in an image as two separate jobs. The output from either job can be shown in the client by displaying dots over the detected structure centroids. Instead of calculating distance between these centroids as a job on the backend, this type of calculation is better fit for the client, once the client has access to the centroids' information. 

The client does not yet display aggregate information such as the average distance. Right now, the distances are visualized as lines between a nucleus centroid and a gland centroid, appearing when either centroid point is moused over.

Since the backend can detect the presence of either nuclei or glands, the database model of `DetectedNucleus` was generalized as `DetectedStructure` and modified to contain information the client would need.


atlascope/core/job_types/gland_detection.py
```diff
@@ -0,0 +1,107 @@
...
+            {
+                key.replace(subset_name, ''): value
+                for key, value in data.items()
+                if subset_name in key
+            },
+        )
+
+    try:
+        input_image = skimage.io.imread(io.BytesIO(open(gland_map_path, 'rb').read()))
+        glands, gland_mask = detect_nuclei(input_image)
```
naglepuff: It seems a bit confusing that the output of a function called `detect_nuclei` would be `glands, gland_mask`. Is this function used by both jobs? I think having a separate function for each structure type would make things more clear.

atlascope/core/job_types/gland_detection.py
```diff
@@ -0,0 +1,107 @@
...
+            additional_gland_attributes = {
+                structure_attribute_to_field_name(attribute): gland[attribute]
+                for attribute in STRUCTURE_ATTRIBUTES
+            }
+            # For small sample image, increase the coordinates of the
+            # centroids to better match the size of the image
+            centroid = Point(
+                x=gland['Identifier.CentroidX'] * 3,
+                y=gland['Identifier.CentroidY'] * 2.5,
+            )
```
naglepuff: I'm a bit confused by what's happening here. What do you mean by "small sample image?" How were the numbers 3 and 2.5 chosen?

web/src/components/PinList.vue
```diff
@@ -55,7 +55,7 @@ import {
   computed, defineComponent, onMounted, Ref, ref, watch,
 } from '@vue/composition-api';
 import store from '../store';
-import { Pin } from '../generatedTypes/AtlascopeTypes';
+import { Dataset, Pin } from '../generatedTypes/AtlascopeTypes';
```
naglepuff: We don't need this import in this component yet. I get an eslint error with it in.

docker-compose.yml
```diff
@@ -1,8 +1,8 @@
 version: '3'
 services:
   postgres:
-    logging:
-      driver: none
+    # logging:
+    #   driver: none
```
naglepuff: Should these lines be uncommented?

web/src/store/index.ts
```diff
@@ -198,7 +212,9 @@ const {
             datasetId: number | undefined;
             result: AxiosResponse;
           }>[] = [];
-          datasets.forEach((dataset) => {
+          datasets.filter(
+            (dataset) => dataset.dataset_type === 'tile_source',
+          ).forEach((dataset) => {
```
naglepuff: I noticed a need for something like this in one of my branches as well. There's a getter above that can be used here instead of `filter`.

```
const { commit, state, dispatch, getters } = rootActionContext(context);
...
getters.tilesourceDatasets.forEach(...)
```

web/src/store/index.ts
```diff
@@ -272,7 +296,26 @@ const {
...
+          && !circlePinDatasetTypes.includes(childDataset.dataset_type)) {
+          return childDataset;
+        }
+        return false;
+      });
+      const circlePins = pins.filter((pin: Pin, index: number) => !isVisualization[index]);
+      const visualizations = isVisualization.filter((vis) => vis !== false);
+
+      commit.setSelectedPins(circlePins);
+      commit.setSelectedVisualizations(visualizations);
```
naglepuff, line 326: There's certainly a lot happening here now that I'm not sure needs to. I'm wondering if these special visualizations really belong to Pins at all. It doesn't seem like the fit the cork board metaphor of pinning something to the larger dataset, since (at least in the nucleus/glands example) the visualizations apply to the whole of the base dataset.

I think having these just be datasets that get handled specially makes more sense than them being Pins. Let me know if you have any thoughts on this. I think it'd be fine to get this in as a first go but I think there are design considerations that would be worth discussing.

web/src/utilities/utiltyFunctions.ts
```diff
@@ -49,3 +51,56 @@ export function opacityForZoomLevel(currentZoom: number, maxZoom: number): numbe
   }
   return maxOpacity - (0.4 * diff);
 }
+
+export function centroidStringToCoords(input: string): Array<number> {
+  const centroidString = input.match(/\(([0-9|.|\s])+\)/);
+  if (!centroidString) return [-1, -1];
```
naglepuff: Could you put this line in brackets? It would make it more readable.
```suggestion
  if (!centroidString) { 
    return [-1, -1];
  }
```

web/src/utilities/utiltyFunctions.ts
```diff
@@ -49,3 +51,56 @@ export function opacityForZoomLevel(currentZoom: number, maxZoom: number): numbe
...
+  const centroid = centroidString[0].slice(1, -1).split(' ').map(
+    (val: string) => parseFloat(val),
+  );
+  return centroid;
+}
+
+export interface NucleusGlandDistance{
+  nucleus: number | undefined;
+  gland: number | undefined;
+  line: Array<Array<number>>;
```
naglepuff: Is this always a 2x2 array? Any thoughts on using `Point[]` as the type?

web/src/utilities/analyticsVis.ts
```diff
@@ -0,0 +1,121 @@
...
+import geo from 'geojs';
+import { GeoJSLayer } from './composableTypes';
+import store from '../store';
+import { centroidStringToCoords, NucleusGlandDistance } from './utiltyFunctions';
+
+interface StructurePoint {
+  x: number;
+  y: number;
+  color: string;
+  struct: DetectedStructure;
```
naglepuff:
```suggestion
  structure: DetectedStructure;
```

I think its more clear to just use the full word here.

naglepuff (changes requested):
Really cool to see this in action. As I mentioned in one of my comments, I'm not entirely convinced that Pins are the best way to hook these datasets up to investigations, but I think it's probably ok for now. I had a some other minor comments as well.

atlascope/core/job_types/gland_detection.py
annehaley: We sent the test image `management/populate/inputs/gland_map.png` into this detection. Our test image was much smaller than the associated dataset image, `HTA9_1_BA_F_ROI-2.ome.tiff` (About 3 times smaller horizontally and 2.5 times smaller vertically)  This simple translation of each centroid was merely to compensate for this size difference and move the detected glands to make them more spacially distributed within  `HTA9_1_BA_F_ROI-2.ome.tiff` (rather than isolated in the top right corner, where there are no nuclei).

Instead of further messing with this, I simply wrote a new `detect_glands` method, inspired by your previous comment. Instead of trying to reuse the nucleus detection on a test image, we can just generate where the glands go by generating n dots in a grid within the size of the input image. 

So this is fixed by [6cbe806](https://github.com/atlascope/atlascope/pull/193/commits/6cbe806fd4df480c24b84b855bcfcf8302a5ef30)

atlascope/core/job_types/gland_detection.py
annehaley: Wrote a new `detect_glands` function in [6cbe806](https://github.com/atlascope/atlascope/pull/193/commits/6cbe806fd4df480c24b84b855bcfcf8302a5ef30)

web/src/components/PinList.vue
annehaley: I left this change out when I rebased and reorganized my commits

docker-compose.yml
annehaley: I left this change out when I rebased and reorganized my commits

web/src/store/index.ts
annehaley: Changed in [5d27ba3](https://github.com/atlascope/atlascope/pull/193/commits/5d27ba37fef2182527ac7f623c2f315f270759a6)

web/src/store/index.ts
annehaley: I agree, and I took a second pass at this design in #201, though the backend of that design remained unchanged. I do think we should return to this topic.

web/src/utilities/analyticsVis.ts
annehaley: Changed in [515411b](https://github.com/atlascope/atlascope/pull/193/commits/515411be71d623eaac177d2da450d4e18aba311b)

naglepuff:
Just a heads up, your most recent force push changed the hash for your recent commits, so they don't match the comments you made on the review. I can manage for now but it would be a good idea to update the comments so they are accurate.

web/src/store/index.ts
naglepuff: Looks like #201 was closed in favor of #204. I'm fine with using pins for these two PRs, but I've opened #206 to try and capture some of my feelings with the current design. At this point I'd like to get this feature merged.

web/src/utilities/utiltyFunctions.ts
annehaley: Changed in [0f3f8b2](https://github.com/atlascope/atlascope/pull/193/commits/0f3f8b2f095924f12167cdf74d4ee27a9ffc0ab4)

web/src/utilities/utiltyFunctions.ts
annehaley: Changed in [39c875f](https://github.com/atlascope/atlascope/pull/193/commits/39c875f4db286a66f66b76328637181902f1ce1c)

annehaley:
> Just a heads up, your most recent force push changed the hash for your recent commits, so they don't match the comments you made on the review. I can manage for now but it would be a good idea to update the comments so they are accurate.

Nice catch, I updated the links in the resolved issues

annehaley:
Apologies for the several force pushes; I'm finding it difficult to merge in changes from main and reconcile the migration files. @mcovalt could you take a look at [3a07b11](https://github.com/atlascope/atlascope/pull/193/commits/3a07b11e36d003f56ca4b08b2c161537ecd5bd2d) and perhaps diagnose why [this test](https://github.com/atlascope/atlascope/runs/7859057521?check_suite_focus=true) failed?

mcovalt:
Looks like the index name was autogenerated to be different when on a different model name. I reordered migrations to have my migration last. Let's see if it works. Ideally, we'd not reorder migrations as this will require dropping the DB (or at least rolling back to 008) to reconcile the new migration sequence. You may want to order them such that your migrations occur after the ones on `main`, but up to you.

annehaley:
Thanks for fixing that. Last question: Is it possible to supply a null/blank value for the fingerprint field? Now we have glands occupying the same model as nuclei, and I'd like to skip the fingerprint part in the `detect_glands` job.

mcovalt:
Looks like the tests are failing due to `fingerprint` not being provided on glands. This exposes a bug in the `CubeField`: a `NOT NULL` constraint should be invalidated immediately, rather than providing what looks like bad data.

Correcting  this behavior in `CubeField` will still result in errors, though. A few options:
- we can allow `null` in the field
- we can just provide a dummy `fingerprint=Cube([-9999]*100)` with glands (probably easiest for now)

In either case, the similar nuclei code assumed `DetectedNuclei` was just nuclei. Expanding its definition to also include glands may break some of the similar nuclei code. I'll keep an eye on this behavior with further development of the similar nuclei features.

annehaley:
Ok, we can provide dummy data there. We do so for a lot of the fields on detected glands, since that data is made up. And yes, combining the two models does complicate the logic for your similarity search. If the design of these two features weren't done simultaneously, we may have had time to come up with a better design concept. Something to consider for Phase 2 perhaps.

naglepuff (approved):
Thanks for getting to all of the feedback and sorting out the migration issues!

waxlamp (commented):
Looks good. Thanks for wrestling with all of those rebasing and migration challenges!

PR #194, opened 2022-07-19
Gland detections

annehaley:
I had this idea while working on #193 and wanted to see what it would look like. When trying to visualize the results of a `average_nucleus_to_gland_distance` job, there are three bits of information the client would need (and thus, that the database must store): the nucleus mask, the gland mask, and the coordinates for the lines visualizing the nearest gland to each nucleus.

But the nucleus mask can already be stored on a previous nucleus detection job. Why not the same for a gland detection job? Right now, that job just invokes the logic of the nucleus detection, but in theory they would be different. And the `average_nucleus_to_gland_distance` job refers to any previous existing detection jobs if they exist, else creating and running them synchronously before accessing their results. 

This way, the client can visualize the `average_nucleus_to_gland_distance` job as only yellow lines when it is visible. Combined with the existing nucleus detection job and gland detection job being visible, the client would render the whole thing without wasting db space with repeated information.

In order to accomplish this, the model for `DetectedNucleus` underwent a rename to `DetectedStructure` to more broadly fit both nuclei and glands (and any future structures).

I know this is a generalization where we probably won't need it, but this organization of job results may help with the visualization step.


waxlamp:
Overall idea seems solid.

Two things come to mind at the moment:
1. If we generalize to "detected structure", how do you name each such dataset so we know what it is (nucleus vs. gland, for example)? (I didn't look closely enough at the code to see if you already answered this question--sorry.)
2. As a general principle, I don't think it's a good idea for a job to kick off another job to fulfill a requirement. Better is to be explicit on both ends: the user needs to run a gland detection job, and then when they run the distance job, they need to explicitly supply both the nucleus detection and gland detection data by name.

annehaley:
1) the `dataset_type` field on the resulting dataset will be either `nucleus_detection` or `gland_detection`.
2) That's a good point, should the distance job return an error if the prereq jobs don't exist yet?

waxlamp:
> 1. the `dataset_type` field on the resulting dataset will be either `nucleus_detection` or `gland_detection`.

Ah, good.

> 2. That's a good point, should the distance job return an error if the prereq jobs don't exist yet?

I suppose. I'm seeing this more like, how can a user even start the job if the prerequisite datasets don't exist yet? That is, how would they specify non-existent datasets in the distance job parameters?

PR #199, opened 2022-08-04
Add endpoint to download datasets

naglepuff:
Datasets can now be downloaded via `/api/v1/datasets/<id>/download`

This endpoint is being added specifically for integrating a local instance of ParaviewGlance, but could be useful for other purposes.

### DIFF
--- a/atlascope/core/rest/endpoints/dataset_endpoints.py
+++ b/atlascope/core/rest/endpoints/dataset_endpoints.py
@@ -75,7 +75,7 @@ class DatasetViewSet(
         dataset = self.get_object()
         try:
             path = dataset.content.path
-        except Exception:
+        except (AttributeError, ValueError):
             raise ValidationError('Dataset has no content')
         if os.path.exists(path):
             response = FileResponse(open(path, 'rb'), content_type='application/octet-stream')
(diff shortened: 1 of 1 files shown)


atlascope/core/rest/endpoints/dataset_endpoints.py
```diff
@@ -57,3 +69,17 @@ def subimage(self, request, pk):
         subimage = original.subimage(**serializer.validated_data)
 
         return Response(DatasetSerializer(subimage).data, status=status.HTTP_201_CREATED)
+
+    @action(detail=True, methods=['GET'], renderer_classes=(PassthroughRenderer,))
+    def download(self, request, pk):
+        dataset = self.get_object()
+        try:
+            path = dataset.content.path
+        except Exception:
```
annehaley: Is there a more precise `Exception` class we can catch here?

annehaley (approved):

atlascope/core/rest/endpoints/dataset_endpoints.py
naglepuff: Yes, updated in e555282

PR #196, opened 2022-07-22
Add `vtk.js` to Atlascope

naglepuff:

PR #200, opened 2022-08-04
Add local instance of Glance and Pin example

naglepuff:
Add a local deployment of [Paraview Glance](https://kitware.github.io/glance/app/) to our `docker` setup. This can be accessed (after building and running `docker-compose up`) at `localhost:9999`.

Add a new dataset and dataset type, and a Pin that connects the new dataset to our deployment of Glance.

Once you re-populate your database with the changes, launch Atlascope. To see the changes in action, navigate to the *Third Party Vis Playground*. Zoom in towards the jaw area of the base dataset until you see a pin appear. Click on this pin to reveal a menu with 2 choices. Click the top choice to launch the local instance of Glance in a new tab with the `Tooth.nrrd` dataset.

To do:
- [x] Add additional click to launch glance/cancel launch


annehaley (changes requested):
I think you may need to change the base on this PR and/or rebase the branch; this seems to hold a lot of the same changes as #196

naglepuff:
> I think you may need to change the base on this PR and/or rebase the branch; this seems to hold a lot of the same changes as #196

I've rebased #196 so that it's more clear that this change should happen first. I think adding this feature is more impactful than the changes in #196, which is still in a status of draft. I still think a choice between Glance and vanilla vtk.js would be interesting to do, but I'm not sure yet how best to do that. I'll revisit that later.

web/src/views/InvestigationDetail.vue
```diff
@@ -44,6 +44,32 @@
...
           />
+          <v-menu
+            v-model="externalVisualizationMenu"
+            :position-x="externalVisualizationMenuX"
+            :position-y="externalVisualizationMenuY"
+            absolute
+            offset-y
+            :close-on-click="false"
+          >
+            <v-list>
```
annehaley: I know this addition is relatively small, but this still may be an opportunity to separate this into its own component. I think that `InvestigationDetail.vue` has grown into a monolith that is difficult to navigate simply because of many small additions like this. It may be too late to refactor the whole thing, but I think it's worthwhile to adopt a philosophy of separation as we go forward.

atlascope/core/management/commands/populate.py
```diff
@@ -109,6 +109,7 @@ def populate_investigations(specs):
         datasets = [Dataset.objects.get(name=name) for name in spec["datasets"]]
         # Only use datasets whose content saved properly
         datasets = [dataset for dataset in datasets if dataset.content]
+        # datasets = [dataset for dataset in datasets if dataset.content]
```
annehaley: This addition can be left out

annehaley (changes requested):

web/src/views/InvestigationDetail.vue
naglepuff: I've created a new component for this context menu in 4d49c64

atlascope/core/management/commands/populate.py
naglepuff: Removed in d4e88c1

annehaley (approved):

PR #204, opened 2022-08-10
Job result vis improvements

annehaley:
The Pins list shown in the Investigation sidebar now has a specialized display for pins that exist to mark the results of a job. This specialized display can include any number of checkboxes which offer available visualizations of the results.

For the job results that are of the type "structure_detection", we offer the following checkboxes:

Centroid locations
Display distances on hover
Each subsequent option does not appear until the prerequisite section encompassing that option is enabled. One additional option exists under "Centroid locations" for nucleus detection jobs:

Color by nearest gland
This option will color each centroid dot by which gland is nearest, resulting in something adjacent to a voronoi diagram.


web/src/components/JobVisOptions.vue
```diff
@@ -0,0 +1,146 @@
...
+import visualize, { visualizeDetectedStructures } from '@/utilities/analyticsVis';
+import {
+  computed, defineComponent, watch,
+} from '@vue/composition-api';
+import store from '../store';
+
+export default defineComponent({
+  props: {
+    pin: {
+      type: Object,
```
naglepuff, line 54: Does `type: Pin` not work here?

web/src/components/JobVisOptions.vue
```diff
@@ -0,0 +1,146 @@
...
+      :key="info"
+    >
+      {{ info }}
+    </v-list-item-subtitle>
+    <div
+      v-for="visOption in visOptions(dataset)"
+      v-show="active"
+      :key="visOption.value"
+      :set="showOptions = visualizationsShown.map(vis => vis.value).includes(visOption.value)"
+      @click="(event) => event.stopPropagation()"
```
naglepuff, line 17: Would just `@click.stop` work or do you need an assignment?

web/src/components/JobVisOptions.vue
```diff
@@ -0,0 +1,146 @@
...
+    const { updateVisualizationsShown, updateVisualizationOption } = store.commit;
+
+    function pinDisplayTitle(pin: Pin) {
+      return pin.description?.replace('Generated by', 'Results of') || 'Job results';
+    }
+    function linkedDataset(pin: Pin) {
+      return currentDatasets.value.find(
+        (dataset: Dataset) => dataset.id === pin.child,
+      );
+    }
```
naglepuff, line 76: I think both `pinDisplayTitle` and `linkedDataset` would be better as computed props. I think you have to pass `props` into `setup` to get access to the Pin in question.

web/src/generatedTypes/AtlascopeTypes.ts
```diff
@@ -452,6 +452,15 @@ export interface Pin {
...
 }
 
+export interface VisOption {
+  value: string,
+  text: string,
+  data: Dataset,
+  visFunc: Function,
+  options: string[],
+  availableOptions: object[],
+}
```
naglepuff, line 462: Was this type automatically generated? I don't remember where it would come from on the back end. If not, re-running the `generateTypes` script will delete this, so I think this type should be defined elsewhere.

web/src/utilities/analyticsVis.ts
```diff
@@ -17,73 +18,96 @@ const defaultStructureColors = {
   gland: 'red',
 };
 
-function drawLines(target: StructurePoint) {
-  const retArray: Array<Array<Array<number>>> = [];
-  const computedLines = computed(() => store.state.nucleiToNearestGlandDistances);
+const colors = distinctColors({
+  count: 50,
```
naglepuff, line 22: Since 20 glands is hard coded into the script for generating glands, would it make sense to use 20 colors instead of 50? Would that help make the colors more visually distinct from one another?

naglepuff (changes requested):

web/src/components/JobVisOptions.vue
annehaley: Unfortunately not, because typescript types aren't sufficient for dictionary values, so we need to use a builtin. else we get this error: `'Pin' only refers to a type, but is being used as a value here.`

annehaley:
Closing in favor of #209

web/src/components/JobVisOptions.vue
annehaley: Do you have a specific way to rewrite that line? I tried using click.stop but it didn't appear to stop the propagation, I think the syntax for that action requires that you pass in a function to run instead of propagation.

web/src/components/JobVisOptions.vue
waxlamp: Indeed, `Pin` in this context is a TypeScript `interface`, and what is needed here is a JavaScript value that fulfills that interface. If you renamed `Pin` (the `interface`) to `PinInterface`, and created a function named `Pin` that yields a value of type `PinInterface`, you could perhaps use `Pin` here in place of `Object`. Not sure about this, though--I'd have to try it and see what the typechecker thinks.

PR #207, opened 2022-08-16
Reconcile conflicting merge operations

naglepuff:
There was an issue with the migration dependency tree due to PR 200 not accounting
for PR 191. Since those two migrations are independant, merging them with django's
manage tool resolved the conflict.


annehaley:
As an additional complication, @mcovalt and I just figured out how to do that with conflicting migrations in #193. Looks like we all wanted to change the db at the same time.

annehaley (approved):

PR #208, opened 2022-08-17
Visualize similar nuclei

mcovalt:
Allows the user to view similar nuclei. See demo video below.

https://user-images.githubusercontent.com/13244041/185230982-2cd83631-3f1b-483e-8570-582321db0f39.mov


web/src/views/InvestigationDetail.vue
```diff
@@ -486,8 +494,57 @@ export default defineComponent({
       clampBoundsX(false);
     }
 
+    const similarNuclei: Ref<Array<DetectedStructureWithColor>> = ref([]);
+    const axios = inject('axios') as AxiosInstance;
+    function colorSimilarNuclei(nuclei: Array<DetectedStructure>) {
+      const colors = ['#FF0000', '#00FF00', '#0000FF', '#FFFF00', '#FF00FF', '#00FFFF'];
```
waxlamp: Don't use these colors. If you need categorical colors (which, see below), pull in D3's categorical colormaps, and use "category10".

Categorical colors to distinguish the markers may be the right way to go, but it could also be interesting to use a linear colormap to encode how similar each nucleus is. That might not be a good idea if you're always showing the 5 most similar (which presumably would all be close to the same color on a linear scale).

web/src/views/InvestigationDetail.vue
```diff
@@ -486,8 +494,57 @@ export default defineComponent({
...
+        'feature',
+        { features: ['marker'] },
+      ) as ReturnType<typeof useGeoJSLayer>;
+      layer.addGeoEventHandler(geoEvents.mouseclick, (event: MouseClickEvent['mouse']) => {
+        const { x, y } = event.geo;
+        if (!rootDataset.value) {
+          return;
+        }
+        const url = `/similar-nuclei/${rootDataset.value.id}/${x}/${y}/`;
+        axios.get(url).then((resp) => {
```
waxlamp: Consider using async/await here.

web/src/components/InvestigationSidebar.vue
```diff
@@ -35,6 +35,12 @@
       <v-tab-item>
         <job-pane />
       </v-tab-item>
+      <v-tab>
+        Similar Nuclei
+      </v-tab>
+      <v-tab-item>
+        <similar-nuclei-pane :similar-nuclei="similarNuclei" />
```
waxlamp: Use PascalCase for custom component declarations (https://v2.vuejs.org/v2/style-guide/?redirect=true#Component-name-casing-in-JS-JSX-strongly-recommended).

web/src/components/InvestigationSidebar.vue
```diff
@@ -49,19 +55,34 @@
...
+  },
+  props: {
+    similarNuclei: {
+      type: Array as PropType<Array<DetectedStructureWithColor>>,
+      required: true,
+    },
   },
 
-  setup() {
+  setup(props) {
```
waxlamp: Why this change?

web/src/components/SimilarNucleiPane.vue
```diff
@@ -0,0 +1,53 @@
...
+} from '../views/InvestigationDetail.vue';
+
+export default defineComponent({
+  props: {
+    similarNuclei: {
+      type: Array as PropType<Array<DetectedStructureWithColor>>,
+      required: true,
+    },
+  },
+  setup(props) {
```
waxlamp: What do style guides recommend about including positional framework method arguments when you don't make use of them?

waxlamp (commented):
This LGTM (though I'm leaving my review as a comment so someone else has to provide a binding review) but I have a couple of ideas you could implement now or in a followup:
1. In addition to showing the 5 most similar nuclei, it might be interesting to also see the 5 *least* similar as well.
2. Change your "RGB" colors out for five colors from "Category10" (talk to me if you want some pairing on this).
3. Consider using a red-blue linear colormap to encode the dissimilar-to-similar range (especially in light of (1) above).
4. It would be great to also see an image of each nucleus in the search results.

Let me know what you think. Looks great!

annehaley (approved):
After Roni's suggestions are addressed, the rest looks good.

web/src/views/InvestigationDetail.vue
mcovalt: Resolved in c865a1abb2f8639637eeece20cfedee49e408328. The new colors look much better!

web/src/views/InvestigationDetail.vue
mcovalt: Awesome. More readable and less nesting. Resolved in a5609958c0259ed93e214de0e88b0f7ae2953355.

web/src/components/InvestigationSidebar.vue
mcovalt: Resolved in 5cc3fd867e4659042d5633a46ade59c4119ebdf8.

web/src/components/InvestigationSidebar.vue
mcovalt: Resolved in 679e31ee02762064b10bd5e5febb71ede6fa683a.

web/src/components/SimilarNucleiPane.vue
mcovalt: Resolved in 679e31ee02762064b10bd5e5febb71ede6fa683a.

PR #209, opened 2022-08-18
Job result vis improvements

annehaley:
This PR is the clean version of #201 and #204, where git rebasing and diverging branches made the history messy.
I tried to address the comments on #204 on this branch, but unfortunately I can't move the conversations here.
I hope this version will still be easier to review.

This PR accomplishes the following:

- Separate the logic for turning on job result visualizations from the logic for turning on pin visibility
- Offer options for job result visualizations as checkboxes that appear when the parent checkbox is active
- Add an option to color nuclei centroids by the nearest gland, resulting in something similar to a voronoi diagram


web/package.json
```diff
@@ -15,6 +15,7 @@
     "@vue/composition-api": "^1.0.0-rc.2",
     "axios": "^0.21.1",
     "direct-vuex": "^0.12.1",
+    "distinct-colors": "^3.0.0",
```
naglepuff: I think @mcovalt is also doing some color-coding stuff in PR #208. Might be worth coordinating what package to use here. I know D3 came up in one of our discussions.

web/src/components/JobVisOptions.vue
```diff
@@ -0,0 +1,149 @@
...
+import visualize, { visualizeDetectedStructures } from '@/utilities/analyticsVis';
+import {
+  computed, defineComponent, watch,
+} from '@vue/composition-api';
+import store from '../store';
+
+export default defineComponent({
+  props: {
+    pin: {
+      type: Object,
```
naglepuff: To continue our discussion from #204, you can do `type: Object as () => Pin`. I think this will be enough for the typescript linter to be able to infer `props.pin`.

web/src/components/JobVisOptions.vue
```diff
@@ -0,0 +1,149 @@
...
+      :key="info"
+    >
+      {{ info }}
+    </v-list-item-subtitle>
+    <div
+      v-for="visOption in visOptions(dataset)"
+      v-show="active"
+      :key="visOption.value"
+      :set="showOptions = visualizationsShown.map(vis => vis.value).includes(visOption.value)"
+      @click="(event) => event.stopPropagation()"
```
naglepuff, line 17: From the conversation on #204, it sounds like `@click.stop=""` doesn't work. 

I'm also curious as to why something like this would need to be added at all to the `<div>` element. Is the problem that the event is bubbling up from child elements? What if you added the `.stop` modifier to the `change` events on the children elements?

web/src/generatedTypes/AtlascopeTypes.ts
```diff
@@ -452,6 +454,15 @@ export interface Pin {
...
 }
 
+export interface VisOption {
+  value: string,
+  text: string,
+  data: Dataset,
+  visFunc: Function,
+  options: string[],
+  availableOptions: object[],
+}
```
naglepuff: Can you re-run `generateTypes`? I see you moved this to `utilities/visualizationTypes.ts`.

naglepuff (changes requested):
Most of these are continuations from #204.

web/package.json
waxlamp: Check out https://github.com/d3/d3-scale-chromatic. This `distinct-colors` library seems neat, but when we're doing any kind of color picking for visualizing, my gut tells me you're better off with the prebaked thought that has gone into D3.

web/src/components/JobVisOptions.vue
annehaley: I tried removing line 17 and changing 22 to `@change.stop="() => updateVisualizationsShown(visOption)"`, as well as 35 to  `@change.stop="() => updateOption(visOption, available.value)"`. But it results in the following behavior:

https://user-images.githubusercontent.com/44912689/185922950-dbf57d02-36f9-4ae5-8227-62862cf6cda8.mov

It seems that the propagation stop has to belong to the direct child of the checkbox content, which is the div in this case. Else the checkbox will interpret that click as a toggle.

web/package.json
annehaley: Ok, I added d3  in [2321cb8](https://github.com/atlascope/atlascope/pull/209/commits/2321cb86ea267fe32e47cfe0ecdcb34c0020d01e) and used schemeCategory10. It doesn't seem to have a scheme with 20 distinct colors, but I suppose it's expected that the colors will loop. With 20 glands present (and therefore 20 color groups), the centroids look like this: 
image
Is this good or should I use a different scheme?

web/src/components/JobVisOptions.vue
annehaley: Nice, I didn't know you could do that. Changed in [edb7bac](https://github.com/atlascope/atlascope/pull/209/commits/edb7bac8abe7d88d0890b1fce959ff5902e9ce26).

web/src/generatedTypes/AtlascopeTypes.ts
annehaley: Re-generated in [3cbf7c3](https://github.com/atlascope/atlascope/pull/209/commits/3cbf7c3d0abd53fdf92ea80fa8fe00ff8e200176)

web/package.json
waxlamp: D3 used to have a `category20` colormap that contained these 10 colors plus darker versions of the same, but it kooks like they got rid of it. But this is a problem inherent to categorical colormapping, that you don't have more than just a handful of colors before you need to repeat. That's fine, especially in this case. (If you try to manufacture your own set of distinct colors, you'll run into the same ultimate problem anyway, of different categories being served by visually close colors.)

web/src/components/JobVisOptions.vue
naglepuff: Hmm, that's interesting. Thanks for testing it out

naglepuff (approved):
LGTM. Something to consider for a follow up would be to make the glands more visually distinct from the nuclei (larger radius, add a border, etc).

Issue #206, opened 2022-08-12
Discussion: Representation of nucleus/gland detection datasets


naglepuff:
PRs #193 and #204 add visualizations for nucleus and gland detection datasets. 

Currently these datasets are linked to investigations via pins, which works well, but I'm not sure if pins are the correct way to accomplish this. I'm wondering if these should just be linked to investigations and have a dedication option in the sidebar.

waxlamp:
Very likely pins are not the right mechanism for these specialized client side visualizations. Actually this is an important topic to discuss in the proposal for Phase 2 (and to focus on in general--what are the right abstractions for the functionality we're trying to develop?--if/when we execute Phase 2).

But I agree that for now it's fine to leave them as is with pins.

PR #212, opened 2022-09-14
Include `3d_volume` pins in display types

waxlamp:
This fixes a bug that was precluding 3d volume pins from being displayed in investigations.

This is further evidence that the pin concept has become overloaded in this codebase (something to address in a possible Phase 2; see #206).

Thanks to @marySalvi for information and context, and @naglepuff for helping me find/fix the error.


waxlamp:
@marySalvi and @naglepuff: I'm ok with a rubber-stamp review, given the low funding for the project at this point.

naglepuff (approved):

